PR #17, opened 2017-07-11
Feature/add compression level parameter

scott-pickthorn:
- adds compression parameter to allow the user to be able to change the compression level
- added command to change enableNygetPackageRestore to set it to true


Test/Install-WhiskeyTool.Tests.ps1
```diff
@@ -251,3 +251,11 @@ Describe 'Install-WhiskeyTool.for non-existent module when version parameter is
     }
 }
 
+Describe 'Install-WhiskeyTool.when set EnableNuGetPackageRestore' {
```
splatteredbits: Change this to use a mock:

```
Mock -CommandName 'Set-Item' -ModuleName 'Whiskey'
It 'should enable NuGet package restore' {
    Assert-MockCalled 'Set-Item' -ModuleName 'Whiskey' -ParameterFilter { $Path -eq 'env:EnableNuGetPackageRestore' }
    Assert-MockCalled 'Set-Item' -ModuleName 'Whiskey' -ParameterFilter { $Value-eq 'true' }
}
```

Test/Invoke-WhiskeyProGetUniversalPackageTask.Tests.ps1
```diff
@@ -1100,3 +1157,26 @@ Describe 'Invoke-WhiskeyProGetUniversalPackageTask.when packaging a directory' {
...
+
+Describe 'Invoke-WhiskeyProGetUniversalPackageTask.when compressionLevel is not included' {
+    GivenARepositoryWithFiles 'one.ps1'
+    WhenPackaging -Paths '*.ps1' -WithWhitelist "*.ps1"
+    ThenPackageShouldbeBeCompressed 'one.ps1' -ExpectedPackageSize 809
+}
+
+Describe 'Invoke-WhiskeyProGetUniversalPackageTask.when a bad compressionLevel is included' {
+    GivenARepositoryWithFiles 'one.ps1'
+    WhenPackagingWithBadCompressionLevelShouldThrowAnError -Paths '*.ps1' -WithWhitelist "*.ps1" -CompressionLevel "this is no good"
```
splatteredbits: Let's do this instead:

```
WhenPackaging -Path '*.ps1' -WithWhitelist '*.ps1' -CompressionLevel 'this is no good'
ThenTaskFails 'not\ a\ valid\ compression\ level'
```

Extract the logic for the `ThenTaskFails` function from the `WhenPackaging` function. Update all the other tests that assumed this would get called to call this assertion explicitly. You'll probably need to create a counter `ThenTaskSucceeds` function (logic extracted from `WhenPackaging`) that all the other tests should use.

Whiskey/Tasks/Invoke-WhiskeyProGetUniversalPackageTask.ps1
```diff
@@ -45,7 +45,10 @@ function Invoke-WhiskeyProGetUniversalPackageTask
     $include = $TaskParameter['Include']
     $exclude = $TaskParameter['Exclude']
     $thirdPartyPath = $TaskParameter['ThirdPartyPath']
-    
+    [int]$compressionLevel = 1
+    if($TaskParameter['CompressionLevel'] -and -not [int]::TryParse($TaskParameter['CompressionLevel'], [ref]$compressionLevel) ){
+        Stop-WhiskeyTask -TaskContext $TaskContext -Message ('"{0}" is not a valid Compression Level. it must be an integer between 0-9.' -f $TaskParameter['CompressionLevel']);
```
splatteredbits: Update error message to include the property name, e.g. `('Property CompressionLevel: ''{0}'' is not a valid compression level. It must be an integer between 0 and 9.'`.

Test/Install-WhiskeyTool.Tests.ps1
scott-pickthorn: done

PR #23, opened 2017-07-14
Add npm registry uri property to node task

scott-pickthorn:
- adding npmRegistryUri property 
- refactored WhiskeyNode Task test file

### DIFF
--- a/Test/Invoke-WhiskeyNodeTask.Tests.ps1
+++ b/Test/Invoke-WhiskeyNodeTask.Tests.ps1
@@ -8,7 +8,7 @@ $startedWithNodeEnv = (Test-Path -Path 'env:NODE_ENV')
 
 $defaultPackageName = 'fubarsnafu'
 
-function WhenBuildIsStarted
+function GivenBuildIsStarted
 {
     [CmdletBinding()]
     param(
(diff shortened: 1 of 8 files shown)


Whiskey/Functions/Invoke-WhiskeyNodeTask.ps1
```diff
@@ -71,7 +73,10 @@ function Invoke-WhiskeyNodeTask
         }
         return
     }
-
+    $npmRegistryUri = $TaskParameter['NpmRegistryUri']
```
splatteredbits: This parameter is required. It should always be provided. We don't want to use `ProGetSession.NpmFeedUri`.

Once you've made `NpmRegistryUri` mandatory (and fixed other issues and failing tests), remove the NpmFeedUri from ProGetSession and remove parameters from all functions that pass it

Test/Invoke-WhiskeyNodeTask.Tests.ps1
```diff
@@ -59,9 +63,19 @@ function Invoke-SuccessfulBuild
     {
         $optionalParams['Clean'] = $True
     }
-
-    Invoke-WhiskeyNodeTask -TaskContext $WithContext -TaskParameter $taskParameter @optionalParams
-
+    if ( $npmRegistryUri )
```
splatteredbits: This should be a `Given` function, e.g. `GivenNpmRegistryUri'. It should set a script-level variable that this function uses.

Whiskey/Functions/Invoke-WhiskeyNodeTask.ps1
scott-pickthorn: Done

PR #41, opened 2017-08-01
Added support for custom OpenCover and ReportGenerator command-line arguments

jeffprice6:


Whiskey/Functions/Invoke-WhiskeyNUnit2Task.ps1
```diff
@@ -47,6 +47,18 @@ function Invoke-WhiskeyNUnit2Task
         $reportGeneratorVersionArg['Version'] = $TaskParameter['ReportGeneratorVersion']
     }
 
+    $openCoverArgs = @()
+    if( $TaskParameter['OpenCoverArgument'] )
+    {
+        $openCoverArgs += $TaskParameter['OpenCoverArgument'] | ForEach-Object {if( $_.StartsWith('-')) { $_ } else { ('-' + $_) } }
```
splatteredbits: Why are you adding `-` to the beginning of the arguments?

jeffprice6: I had previously used @() instead of @{} for the custom parameters. I've updated this so and it's no splatting the values correctly..

Whiskey/Functions/Invoke-WhiskeyNUnit2Task.ps1
```diff
@@ -46,7 +46,10 @@ function Invoke-WhiskeyNUnit2Task
     {
         $reportGeneratorVersionArg['Version'] = $TaskParameter['ReportGeneratorVersion']
     }
-
+    
+    $openCoverArgs = $TaskParameter['OpenCoverArgument']
```
splatteredbits: These should be arrays of values that get passed to the command as-is. Using splatting, they can only pass/use in parameters that begin with `-`. 

Pass the args directly to the commands without splatting. 

When displaying them as verbose messages, separate them by spaces, e.g. `$openCoverArgs -join ' '`.

splatteredbits:
Awesome! Will you increment the version number in the whiskey.yml and Whiskey.psd1 files to 0.12.0? Will you update the release notes in the Whiskey.psd1 file? Then I'll merge it in.

PR #56, opened 2017-08-23
Feature/46

splatteredbits:


Whiskey/Tasks/Invoke-WhiskeyPipelineTask.ps1
```diff
@@ -0,0 +1,105 @@
+ function Invoke-PipelineTask
```
jeffprice6: Per the other tasks in Whiskey, should this be named `Invoke-WhiskeyPipelineTask`?

splatteredbits: There is already a function named `Invoke-WhiskeyPipeline`, so I had to name this `Invoke-WhiskeyPipelineTask`. `:(`

splatteredbits: Oh, hah! I just noticed the typo. I didn't look at the code. Yes. This should be named `Invoke-WhiskeyPipelineTask`.

PR #58, opened 2017-08-24
Feature/issue 49

### LINKED ISSUE #49: Specify version of NuGet to use in tasks that use NuGet
These tasks use `NuGet.exe`. Update them to behave like the `NuGetRestore` task so users can specify what version of NuGet to use.

* `MSBuild`
* `NuGetPush`
* `NuGetPack`

jlariono:
Fixes #49

### DIFF
--- a/Whiskey/Functions/Invoke-WhiskeyNuGetPush.ps1
+++ b/Whiskey/Functions/Invoke-WhiskeyNuGetPush.ps1
@@ -13,14 +13,15 @@ function Invoke-WhiskeyNuGetPush
 
         [Parameter(Mandatory=$true)]
         [string]
-        $ApiKey
+        $ApiKey,
+
+        [string]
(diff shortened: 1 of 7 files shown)


Whiskey/Functions/Invoke-WhiskeyNuGetPush.ps1
```diff
@@ -13,14 +13,15 @@ function Invoke-WhiskeyNuGetPush
 
         [Parameter(Mandatory=$true)]
         [string]
-        $ApiKey
+        $ApiKey,
+
+        [string]
```
splatteredbits: Make this a mandatory parameter. This function exists so we can mock it during testing. It is internal and only used in one place.

jlariono: Fixed.

PR #59, opened 2017-08-28
Created Process whiskey task

### LINKED ISSUE #52: Create Process task
Create a process task that starts an external process.

jlariono:
Fixes #52

### DIFF
--- a/Whiskey/Tasks/Invoke-WhiskeyProcess.ps1
+++ b/Whiskey/Tasks/Invoke-WhiskeyProcess.ps1
@@ -2,42 +2,57 @@ function Invoke-WhiskeyProcess
 {
     <#
     .SYNOPSIS
-    Starts a process with given arguments
+    Runs a process\executable.
     
     .DESCRIPTION
-    The `Process` task runs an executable located at a given `Path`, relative to your `whiskey.yml` file. Additionally, you may specify a list of `Argument` and a `WorkingDirectory` in which to start the process.
-    
-    The exit code for the process must be '0' or one of `SuccessExitCode` specified, otherwise the build will be failed.
-    
-    ## Properties
+    The `Process` task runs a process/executable. Specify the path to the executable to run with the task's `Path` property. The `Path` can be the name of an executable that can be found in the `PATH` environment variable, a path relative to your `whiskey.yml` file's directory, or an absolute path.
 
-    ### Mandatory
-    * `Path`: path to the executable to run, relative to `whiskey.yml`
-    
-    ### Optional
-    * `Argument`: a list of arguments to be passed to the executable
-    * `SuccessExitCode`: a list of exit codes that indicate the process ran successfully
-    * `WorkingDirectory`: the directory in which to start the process
+    The task will fail if the process returns a non-zero exit code. Use the `SuccessExitCode` property to configure the task to interpret other exit codes as "success". 
 
-    ## Examples
+    Pass arguments to the process via the `Argument` property. The `Process` task uses PowerShell's `Start-Process` cmdlet to run the process, so that arguments will be passes as-is, with no escaping. YAML strings, however, are usually single-quoted (e.g. `'Value'`) or double-quoted (e.g. `"Value"`). If you're using a single quoted string and need to insert a single quote, escape it by using two single quotes, e.g. `'escape: '''` is converted to `escape '`. If you're using a double-quoted string and need to insert a double quote, escape it with `\`, e.g. `"escape: \""` is converted to `escape: "`. YAML supports other escape sequences in double-quoted strings. The full list of escape sequences is in the [YAML specification](http://yaml.org/spec/current.html#escaping in double quoted style/).
 
-    ### Example 1
+    By default, the executable is run from your `whiskey.yml` file's directory (i.e. the build root). Change the working directory with the `WorkingDirectory` property.
 
-        BuildTasks:
-        - Process:
-            Path: NCrunch.exe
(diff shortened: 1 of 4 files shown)


Whiskey/Tasks/Invoke-WhiskeyProcess.ps1
```diff
@@ -0,0 +1,118 @@
+function Invoke-WhiskeyProcess
+{
+    <#
+    .SYNOPSIS
+    Starts a process with given arguments
+    
+    .DESCRIPTION
```
splatteredbits: Change to:

```
The `Process` task runs a process/executable. Specify the path to the executable to run with the task's `Path` property. The `Path` can be the name of an executable that can be found in the `PATH` environment variable, a path relative to your `whiskey.yml` file's directory, or an absolute path.

The task will fail if the process returns a non-zero exit code. Use the `SuccessExitCode` property to configure the task to interpret other exit codes as "success". 

Pass arguments to the process via the `Argument` property. The `Process` task uses PowerShell's `Start-Process` cmdlet to run the process, so that arguments will be passes as-is, with no escaping. YAML strings, however, are usually single-quoted (e.g. `'Value'`) or double-quoted (e.g. `"Value"`). If you're using a single quoted string and need to insert a single quote, escape it by using two single quotes, e.g. `'escape: '''` is converted to `escape '`. If you're using a double-quoted string and need to insert a double quote, escape it with `\`, e.g. `"escape: \""` is converted to `escape: "`. YAML supports other escape sequences in double-quoted strings. The full list of escape sequences is in the [YAML specification](http://yaml.org/spec/current.html#escaping in double quoted style/).

By default, the executable is run from your `whiskey.yml` file's directory (i.e. the build root). Change the working directory with the `WorkingDirectory` property.

# Properties

* `Path` (*mandatory*): the path to the executable to run. This can be the name of an executable if it is in your PATH environment variable, a path relative to the `whiskey.yml` file, or an absolute path.
* `Argument`: a list of arguments to pass to the executable. Read the documentation above for notes on how to properly escape arguments.
* `WorkingDirectory`: the directory the executable will run in/from. By default, this is the build root, i.e. the `whiskey.yml` file's directory.
* `SuccessExitCode`: a list of exit codes that the `Process` task should interpret to mean the process exited successfully. The default is `0`.

# Examples

## Example 1

        BuildTasks:
        - Process:
            Path: cmd.exe
            Argument:
            - /C
            - dir C:\

This example demonstrates how to call an executable whose arguments have to be quoted a specific way. In this case, we're using `cmd.exe` to get a directory listing of the `C:\` directory. This example will run `cmd.exe /C dir C:\.

## Example 2

        BuildTasks:
        - Process:
            Path: robocopy.exe
            Argument:
            - C:\Source
            - C:\Desitination
            - /MIR    
            SuccessExitCode:
            - 0
            - 1
            - 2
            - 3
            - 4
            - 5
            - 6
            - 7

This example demonstrates how to configure the `Process` task to fail when an executable can return multiple success exit codes. In this case, `robocopy.exe` can return any value less than 8 to report a successful copy.
```

Whiskey/Tasks/Invoke-WhiskeyProcess.ps1
```diff
@@ -58,11 +73,22 @@ function Invoke-WhiskeyProcess
...
+            BuildTasks:
+            - Process:
+                Path: cmd.exe
+            
+        ')
     }
 
     $processPath = ''
-    if ( Test-Path -Path $path -PathType Leaf )
+    $pathRelativeToBuildRoot = Join-Path -Path $TaskContext.BuildRoot -ChildPath $path
```
splatteredbits: This is getting hard for me to understand. Does this work (replacing lines 85 to 104)?

```
if( -not [IO.Path]::IsPathRooted($path) )
{
    $path = Join-Path -Path $TaskContext.BuildRoot -ChildPath $path
}

if( (Test-Path -Path $path -PathType Leaf) )
{
    $path = $path | Resolve-Path | Select-Object -ExpandProperty 'ProviderPath'
}
else
{
    $path = $TaskParameter['Path']
    if( -not (Get-Command -Name $path -CommandType Application -ErrorAction Ignore) )
    {
        Stop-WhiskeyTask -TaskContext $TaskContext -Message ('Executable ''{0}'' does not exist. We checked if the executable is at that path on the file system and if it is in your PATH environment variable.' -f $path)
    }
}
```

jlariono: Much more concise, I like it.

Whiskey/Tasks/Invoke-WhiskeyProcess.ps1
```diff
@@ -0,0 +1,140 @@
...
+                - 4
+                - 5
+                - 6
+                - 7
+
+    This example demonstrates how to configure the `Process` task to fail when an executable can return multiple success exit codes. In this case, `robocopy.exe` can return any value less than 8 to report a successful copy.
+    #>      
+
+    [CmdletBinding()]
+    [Whiskey.Task("Process")]
```
splatteredbits: What do you think about renaming this task to `Exec`? I was never quite comfortable with `Process`. I think I like `Exec` better. What do you think?

jlariono: I agree, Exec is much more clear to the user that this task can run any arbitrary executable rather than a defined set of Processes.

I'll go ahead and make the renaming changes.

PR #81, opened 2017-09-20
feature/delete task

splatteredbits:
Created a Delete task for deleting files/directories.

### DIFF
--- a/Whiskey/Tasks/Remove-WhiskeyItem.ps1
+++ b/Whiskey/Tasks/Remove-WhiskeyItem.ps1
@@ -1,0 +1,71 @@
+function Remove-WhiskeyItem
+{
+    <#
+    .SYNOPSIS
+    Deletes files or directories.
+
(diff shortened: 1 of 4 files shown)


Whiskey/Tasks/Remove-WhiskeyItem.ps1
```diff
@@ -0,0 +1,71 @@
...
+
+        [Parameter(Mandatory=$true)]
+        [hashtable]
+        $TaskParameter
+    )
+
+    Set-StrictMode -Version 'Latest'
+    Use-CallerPreference -Cmdlet $PSCmdlet -SessionState $ExecutionContext.SessionState
+
+    foreach( $path in $TaskParameter['Path'] )
```
jlariono: You listed `Path` as mandatory in the docs but looks like the task will quietly continue and do nothing if `Path` isn't specified. Maybe that was intentional though?

PR #82, opened 2017-09-20
'npm prune' now occurs in the 'PublishNodeModule' task

jlariono:
* 'npm prune' now occurs in the 'PublishNodeModule' task
* 'npm prune' will use the locally installed version of npm if it exists

### DIFF
--- a/Whiskey/Tasks/Publish-WhiskeyNodeModule.ps1
+++ b/Whiskey/Tasks/Publish-WhiskeyNodeModule.ps1
@@ -42,8 +42,7 @@ function Publish-WhiskeyNodeModule
     Set-StrictMode -Version 'Latest'
     Use-CallerPreference -Cmdlet $PSCmdlet -SessionState $ExecutionContext.SessionState
 
-    $buildRoot = $TaskContext.BuildRoot
-    $workingDir = $buildRoot
+    $workingDir = $TaskContext.BuildRoot
     if($TaskParameter.ContainsKey('WorkingDirectory'))
     {
         $workingDir = $TaskParameter['WorkingDirectory'] | Resolve-WhiskeyTaskPath -TaskContext $TaskContext -PropertyName 'WorkingDirectory'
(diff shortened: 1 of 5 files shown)


Whiskey/Tasks/Publish-WhiskeyNodeModule.ps1
```diff
@@ -124,12 +120,29 @@ function Publish-WhiskeyNodeModule
...
+            & $nodePath $npmPath prune --production --no-color
+        }
+        
+        if ($LASTEXITCODE -ne 0)
+        {
+            Stop-WhiskeyTask -TaskContext $TaskContext -Message ('NPM command ''npm prune'' failed with exit code ''{0}''.' -f $LASTEXITCODE)
+        }
+        
+        # local version of npm gets removed by 'npm prune', so call Get-WhiskeyNPMPath to download it again so we can also use the desired version of npm for publishing
+        $npmPath = Get-WhiskeyNPMPath -NodePath $nodePath -ApplicationRoot $workingDir
```
splatteredbits: Do you need do call `Get-WhiskeyNPMPath` twice?

splatteredbits:
Oh, and don't forget to update the release notes in Whiskey.psd1.

Whiskey/Tasks/Publish-WhiskeyNodeModule.ps1
jlariono: If using a local npm then npm will prune itself with the `npm prune` call just before this line. So we need to call `Get-WhiskeyNPMPath` again to have that function re-download the local npm so we can also run `npm publish` with that version of npm. We can avoid this if we use the global version of npm to do the `npm publish` after we have run `npm prune` with the local npm. But that would mean that publishing wouldn't use the version of npm that the rest of the build is using. What do you think?

splatteredbits: That makes sense. Leave it as-is.

jlariono:
Added release notes to Whiskey.psd1.

PR #88, opened 2017-09-28
Feature/npm prune

jlariono:
Added an `NpmPrune` task.


Test/Invoke-WhiskeyNpmPrune.Tests.ps1
```diff
@@ -0,0 +1,259 @@
...
+    It 'should not write any errors' {
+        $Global:Error | Should -BeNullOrEmpty
+    }
+
+    It 'should not fail' {
+        $failed | Should -Be $false
+    }
+}
+
+Describe 'NpmPrune.when running in Initialize mode' {
```
splatteredbits: Add a `Then` assertion that checks if the task really initialized, i.e. if Node.js was installed. I would probably mock out the call to 'Install-WhiskeyNodeJs`.

jlariono: Good catch. Fixed.

PR #91, opened 2017-10-06
Updated `Resolve-WhiskeyVariable` to correctly resolve well-known var…

jeffprice6:
…iables containing values of '0'

### DIFF
--- a/Test/Resolve-WhiskeyVariable.Tests.ps1
+++ b/Test/Resolve-WhiskeyVariable.Tests.ps1
@@ -251,17 +251,6 @@ Describe 'Resolve-WhiskeyVariable.when using well-known variables' {
     ThenValueIs $expectedValue
 }
 
-Describe 'Resolve-WhiskeyVariable.when using well-known variable with a value of ''0''' {
-    Init
-    WhenResolving '$(WHISKEY_BUILD_NUMBER)'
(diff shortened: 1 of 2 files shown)


Test/Resolve-WhiskeyVariable.Tests.ps1
```diff
@@ -251,6 +251,17 @@ Describe 'Resolve-WhiskeyVariable.when using well-known variables' {
     ThenValueIs $expectedValue
 }
 
+Describe 'Resolve-WhiskeyVariable.when using well-known variable with a value of ''0''' {
+    Init
```
splatteredbits: Write this test case like this:

```
Init
GivenVariable 'SOMETHING' '0'
WhenResolving '$(SOMETHING)'
ThenValueIs '0'
```

jeffprice6: I had it like this, but the test passed even before I made the changes to the function. I can try to mess around with the data types being passed to try to get it to fail.

splatteredbits: `'0'` is a string, so has a value. Change the value to an actual `0`, e.g.

```
GivenVariable 'SOMETHING' 0
```

PR #118, opened 2017-12-01
0.23.0

splatteredbits:
Every task now gets its own temp directory under `.output`. This is a better place since virus scan exclusions are usually made on an entire drive/directory and I don't want to fight to exclude the build user's temp directory.

Removed the default .hg, .git, and obj exclusion items so that if/when someone actually wants to include them, they can.

### DIFF
--- a/Test/Invoke-WhiskeyTask.Tests.ps1
+++ b/Test/Invoke-WhiskeyTask.Tests.ps1
@@ -388,6 +388,42 @@ function ThenTaskRanWithoutParameter
     }
 }
 
+function ThenTempDirectoryCreated 
+{
+    param(
+        $TaskName
+    )
+
+    $expectedTempPath = Join-Path -Path $context.OutputDirectory -ChildPath ('Temp.{0}.' -f $TaskName)
+    $expectedTempPathRegex = '^{0}[a-z0-9]{{8}}\.[a-z0-9]{{3}}$' -f [regex]::escape($expectedTempPath)
+    It ('should create a task-specific temp directory') {
+        Assert-MockCalled -CommandName 'New-Item' -ModuleName 'Whiskey' -ParameterFilter { 
+            #$DebugPreference = 'Continue'
+            Write-Debug ('Path  expected  {0}' -f $expectedTempPathRegex)
+            Write-Debug ('      actual    {0}' -f $Path)
+            $Path -match $expectedTempPathRegex }
+        Assert-MockCalled -CommandName 'New-Item' -ModuleName 'Whiskey' -ParameterFilter { $Force }
+        Assert-MockCalled -CommandName 'New-Item' -ModuleName 'Whiskey' -ParameterFilter { 
+            #$DebugPreference = 'Continue'
+            Write-Debug ('ItemType  expected  {0}' -f 'Directory')
+            Write-Debug ('          actual    {0}' -f $ItemType)
+            $ItemType -eq 'Directory' }
+    }
+}
+
+function ThenTempDirectoryRemoved 
+{
+    param(
+        $TaskName
+    )
+    
+    $expectedTempPath = Join-Path -Path $context.OutputDirectory -ChildPath ('Temp.{0}.*' -f $TaskName)
+    It ('should delete task-specific temp directory') {
+        $expectedTempPath | Should -Not -Exist
+        $context.Temp | Should -Not -Exist
+    }
+}
+
 function ThenThrewException
 {
     param(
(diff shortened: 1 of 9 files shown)


splatteredbits:
@jlariono @jeffprice6 Can one of you review and approve this change? Thanks.

Whiskey/Functions/Invoke-WhiskeyTask.ps1
```diff
@@ -50,11 +50,20 @@ function Invoke-WhiskeyTask
             $result = 'FAILED'
             try
             {
+                $TaskContext.Temp = Join-Path -Path $TaskContext.OutputDirectory -ChildPath ('Temp.{0}.On{1}.{2}' -f $Name,$EventName,[IO.Path]::GetRandomFileName())
+                if( -not (Test-Path -Path $TaskContext.Temp -PathType Container) )
+                {
+                    New-Item -Path $TaskContext.Temp -ItemType 'Directory' -Force
```
jlariono: Should add an `Out-Null` here.

Whiskey/Functions/Invoke-WhiskeyTask.ps1
```diff
@@ -244,11 +253,20 @@ function Invoke-WhiskeyTask
     $result = 'FAILED'
     try
     {
+        $TaskContext.Temp = Join-Path -Path $TaskContext.OutputDirectory -ChildPath ('Temp.{0}.{1}' -f $task.Name,[IO.Path]::GetRandomFileName())
```
jlariono: Here you used `$task.Name` but above on line 53 you used `$Name`. Should probably pick one to be consistent.

Test/Invoke-WhiskeyTask.Tests.ps1
```diff
@@ -388,6 +388,42 @@ function ThenTaskRanWithoutParameter
     }
 }
 
+function ThenTempDirectoryCreated 
+{
+    param(
+        $TaskName
+    )
+
+    $expectedTempPath = Join-Path -Path $context.OutputDirectory -ChildPath ('Temp.{0}.' -f $TaskName)
```
jlariono: I *think* for the regex to be correct you need to escape the `.` character in `Temp.{0}.` as well as any `.` characters in `$TaskName`.

Test/Invoke-WhiskeyTask.Tests.ps1
```diff
@@ -388,6 +388,42 @@ function ThenTaskRanWithoutParameter
...
+function ThenTempDirectoryCreated 
+{
+    param(
+        $TaskName
+    )
+
+    $expectedTempPath = Join-Path -Path $context.OutputDirectory -ChildPath ('Temp.{0}.' -f $TaskName)
+    $expectedTempPathRegex = '^{0}[a-z0-9]{{8}}\.[a-z0-9]{{3}}$' -f [regex]::escape($expectedTempPath)
+    It ('should create a task-specific temp directory') {
+        Assert-MockCalled -CommandName 'New-Item' -ModuleName 'Whiskey' -ParameterFilter {
```
jlariono: I've always been curious why you do separate `Assert-MockCalled` for each individual parameter instead of checking all the parameters with a single `Assert-MockCalled`. Doesn't doing them separately leave it possible for this test to pass even if `New-Item` doesn't contain all the parameters in a single call?
e.g.
```
New-Item -Path $expectedTempPathRegex -ItemType 'File'
New-Item -Path 'someOtherPath' -ItemType 'Directory'
```

Test/Invoke-WhiskeyTask.Tests.ps1
splatteredbits: On the next line, you'll see I'm using `[regex]::escape` to escape all those unfriendly characters.

Test/Invoke-WhiskeyTask.Tests.ps1
splatteredbits: Because I can't debug into the parameter filter blocks. I used to combine everything. When the assertion failed, I had to write a crap load of `Write-Debug` statements to figure out which parameter was invalid. On separate lines, I can look at the Pester failure message, see what line failed, know what parameter was wrong, and usually I can figure out why it is broken without having to add any `Write-Debug` statements. You can see here, however, that this time I couldn't. :(

Whiskey/Functions/Invoke-WhiskeyTask.ps1
splatteredbits: Fixed.

Whiskey/Functions/Invoke-WhiskeyTask.ps1
splatteredbits: Fixed.

splatteredbits:
@jlariono Ready for another review.

Test/Invoke-WhiskeyTask.Tests.ps1
jlariono: Ah yes, I skipped over that.

Test/Invoke-WhiskeyTask.Tests.ps1
jlariono: Oh that makes sense. Thanks for the explanation.

PR #121, opened 2017-12-13
Bugfix/search when searching not required

splatteredbits:
Only search module repositories if version has wildcard characters.


splatteredbits:
Now that `Resolve-WhiskeyPowerShellModule` returns module object instead of a version, this fix doesn't work. Closing.

PR #123, opened 2017-12-19
VSDEVOPS-361 - Added `NspVersion` property to NspCheck task and remov…

jeffprice6:
…ed pinned (2.7.0) version of NSP

### DIFF
--- a/Whiskey/Tasks/Invoke-WhiskeyNspCheck.ps1
+++ b/Whiskey/Tasks/Invoke-WhiskeyNspCheck.ps1
@@ -16,6 +16,7 @@ function Invoke-WhiskeyNspCheck
 
     # * `NpmRegistryUri` (mandatory): the uri to set a custom npm registry.
     # * `WorkingDirectory`: the directory where the `package.json` exists. Defaults to the directory where the build's `whiskey.yml` file was found. Must be relative to the `whiskey.yml` file.
+    # * `NspVersion`: the version of NSP to install and utilize for security checks. Defaults to the latest stable version of NSP.
 
     # Examples
 
@@ -35,6 +36,15 @@ function Invoke-WhiskeyNspCheck
             WorkingDirectory: app
     
     This example will run `node.exe nsp check` against the modules listed in the `package.json` file that is located in the `(BUILD_ROOT)\app` directory.
+
+    ## Example 3
+
+        BuildTasks:
+        - NspCheck:
+            NpmRegistryUri: "http://registry.npmjs.org"
+            NspVersion: 2.7.0
+    
+    This example will run `node.exe nsp check` by installing and running NSP version 2.7.0.
     #>
 
     [Whiskey.Task("NspCheck", SupportsClean=$true, SupportsInitialize=$true)]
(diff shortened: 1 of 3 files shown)


Whiskey/Tasks/Invoke-WhiskeyNspCheck.ps1
```diff
@@ -113,10 +123,22 @@ function Invoke-WhiskeyNspCheck
 
         Write-Timing -Message 'Running NSP security check'
 
-        $output = Invoke-Command -NoNewScope -ScriptBlock {
-            & $nodePath $nspPath 'check' '--output' 'json' 2>&1 |
-                ForEach-Object { if( $_ -is [Management.Automation.ErrorRecord]) { $_.Exception.Message } else { $_ } }
+        if( !$nspVersion -or $nspVersion -gt (ConvertTo-WhiskeySemanticVersion -InputObject '2.7.0') )
```
jlariono: How about if we use this `if` statement to simply set an `$outputParam` variable to either `--reporter` or `--output`? That way we don't repeat the `Invoke-Command` logic twice.

jeffprice6: Updated this logic. I added the -ArgumentList parameter to the Invoke-Command so that our Pester tests can see the value being passed. I thought of creating the entire scriptblock outside of the function (so it would resolve the variables for Pester) and passing that in, but chose to go this route..

jlariono: Oh good idea, I hadn't thought of doing it that way.

Whiskey/Tasks/Invoke-WhiskeyNspCheck.ps1
```diff
@@ -91,9 +101,19 @@ function Invoke-WhiskeyNspCheck
     }
 
     Write-Timing -Message 'Installing NSP'
-    $nspModuleRoot = Install-WhiskeyNodeModule -Name 'nsp' -Version '2.7.0' -ApplicationRoot $workingDirectory -RegistryUri $npmRegistryUri -ForDeveloper:$TaskContext.ByDeveloper
+
+    if( $TaskParameter['NspVersion'] )
```
jlariono: I think this property should actually be named just `Version` without the `Nsp` to be consistent with the other tasks that have a `Version` property.

jeffprice6: Agreed, I've updated to match the standard convention!

Whiskey/Tasks/Invoke-WhiskeyNspCheck.ps1
```diff
@@ -113,10 +133,21 @@ function Invoke-WhiskeyNspCheck
 
         Write-Timing -Message 'Running NSP security check'
 
+        $formattingArg = '--output'
+        if( !$nspVersion -or $nspVersion -gt (ConvertTo-WhiskeySemanticVersion -InputObject '2.7.0') )
```
splatteredbits: This should be `$nspVersion -ge (ConvertTo-WhiskeySemanticVersion -InputObject '3.0.0')`. Theoretically, they could release a 2.x or 2.7.x that continues to use the `--output` flag.

jeffprice6: Great point. I'll be updating the NSP check task to support the global WorkingDirectory property shortly. I'll commit this change too when I push it up.. :)

PR #129, opened 2018-01-03
Updated `Exec` task to support a single-line syntax..

jeffprice6:
..to define executable path and arguments.

### DIFF
--- a/Test/Invoke-WhiskeyExec.Tests.ps1
+++ b/Test/Invoke-WhiskeyExec.Tests.ps1
@@ -105,13 +105,14 @@ function GivenTaskDefaultProperty
         $Property
     )
 
-    $script:defaultProperty = Select-String -InputObject $Property -Pattern '([^\s"'']+)|"([^"]*)"|''([^'']*)''' -AllMatches
+    $regExMatches = Select-String -InputObject $Property -Pattern '([^\s"'']+)|"([^"]*)"|''([^'']*)''' -AllMatches
+    $defaultProperty = @($regExMatches.Matches.Groups | Where-Object { $_.Name -ne '0' -and $_.Success -eq $true } | Select-Object -ExpandProperty 'Value')
 
-    $script:path = $defaultProperty.Matches | Where-Object { $_.Index -eq 0 } | Select-Object -ExpandProperty 'Groups' |
-        Where-Object { $_.Name -ne '0' -and $_.Success -eq $true } | Select-Object -ExpandProperty 'Value'
-
-    $script:argument = $defaultProperty.Matches | Where-Object { $_.Index -ne 0 } | Select-Object -ExpandProperty 'Groups' |
-        Where-Object { $_.Name -ne '0' -and $_.Success -eq $true } | Select-Object -ExpandProperty 'Value'
+    $script:path = $defaultProperty[0]
+    if( $defaultProperty.Count -gt 1 )
+    {
+        $script:argument = $defaultProperty[1..($defaultProperty.Count - 1)]
+    }
 }
 
 function WhenRunningExecutable
--- a/Test/Invoke-WhiskeyExec.Tests.ps1
+++ b/Test/Invoke-WhiskeyExec.Tests.ps1
@@ -9,6 +9,7 @@ $failed = $false
 $path = $null
 $successExitCode = $null
 $workingDirectory = $null
+$defaultProperty = $null
 
 function Get-BuildRoot
 {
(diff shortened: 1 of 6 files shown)


Whiskey/Tasks/Invoke-WhiskeyExec.ps1
```diff
@@ -11,6 +11,8 @@ function Invoke-WhiskeyExec
 
     Pass arguments to the executable via the `Argument` property. The `Exec` task uses PowerShell's `Start-Process` cmdlet to run the executable, so that arguments will be passes as-is, with no escaping. YAML strings, however, are usually single-quoted (e.g. `'Value'`) or double-quoted (e.g. `"Value"`). If you're using a single quoted string and need to insert a single quote, escape it by using two single quotes, e.g. `'escape: '''` is converted to `escape '`. If you're using a double-quoted string and need to insert a double quote, escape it with `\`, e.g. `"escape: \""` is converted to `escape: "`. YAML supports other escape sequences in double-quoted strings. The full list of escape sequences is in the [YAML specification](http://yaml.org/spec/current.html#escaping in double quoted style/).
 
+    The `Exec` task supports a simplified single line syntax to define the `Path` and optional `Arguments` properties. Anything enclosed by single-quote or double-quote characters are treated as an individual path or argument. Otherwise, white-space is the default delimiter separating items.
```
splatteredbits: Give an example.

```
BuildTasks:
- Exec: powershell.exe -NonInteractive -NoProfile -ExecutionPolicy Bypass $PSVersionTAble
```

Test/Invoke-WhiskeyPipeline.Tests.ps1
```diff
@@ -297,6 +297,20 @@ BuildTasks:
     }
 }
 
+Describe 'Invoke-WhiskeyPipeline.when task has a default property' {
```
splatteredbits: Add some tests with some real-world command line examples, e.g.

```
robocopy "C:\Some File" 'D:\Some Other\File' /MIR
```

We want tests to ensure the YAML is getting parsed correctly.

Test/Invoke-WhiskeyExec.Tests.ps1
```diff
@@ -97,6 +99,22 @@ function GivenSuccessExitCode
     $script:successExitCode = $SuccessExitCode
 }
 
+function GivenTaskDefaultProperty
+{
+    param(
+        $Property
+    )
+
+    $regExMatches = Select-String -InputObject $Property -Pattern '([^\s"'']+)|"([^"]*)"|''([^'']*)''' -AllMatches
```
splatteredbits: Don't do this parsing. If other assertions needs the path/argument values, then update those assertions to take the expected path/arguments as parameters and have your tests pass them. There could be a bug hidden here. Explicit tests make it easier to spot when expections/assertions/preconditions don't line up.

jeffprice6: Good call, I've updated it so it's actually testing something..

Test/Invoke-WhiskeyPipeline.Tests.ps1
jeffprice6: I've added a couple tests to ensure the property gets passed as defined.

Whiskey/Tasks/Invoke-WhiskeyExec.ps1
jeffprice6: I previously added an Example #3. Did you have something else in mind or just overlooked? :)

Test/Invoke-WhiskeyExec.Tests.ps1
```diff
@@ -256,7 +258,7 @@ Describe 'Invoke-WhiskeyExec.when running an executable with multiple arguments'
 Describe 'Invoke-WhiskeyExec.when utilizing default task property to define executable and arguments' {
```
splatteredbits: This is what I'd like to see. Instead of passing the expected arguments to a `Given` function, pass them to an assertion.

```
Init
GivenExecutableFile 'executable.bat' 'exit 0'
GivenTaskDefaultProperty -Property 'executable.bat Arg1 Arg2 "Arg 3" ''Arg 4'''
WhenRunningExecutable
ThenExecutableRan 'executable.bat'
ThenSpecifiedArgumentsWerePassed 'Arg1','Arg2','"Arg 3"',"'Arg 4'"
```

jeffprice6: Updated tests to align with these suggestions.

PR #139, opened 2018-02-08
`New-WhiskeySemanticVersion` could not read `Version` property from .…

jeffprice6:
…csproj files if they contained a default xml namespace

### DIFF
--- a/Whiskey/Functions/New-WhiskeySemanticVersion.ps1
+++ b/Whiskey/Functions/New-WhiskeySemanticVersion.ps1
@@ -128,11 +128,18 @@ function New-WhiskeySemanticVersion
                 return
             }
 
-            $csprojVersion = $csprojXml.SelectNodes('/Project/PropertyGroup/Version') | Select-Object -ExpandProperty '#text'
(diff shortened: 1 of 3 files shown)


Whiskey/Functions/New-WhiskeySemanticVersion.ps1
```diff
@@ -128,11 +128,18 @@ function New-WhiskeySemanticVersion
                 return
             }
 
-            $csprojVersion = $csprojXml.SelectNodes('/Project/PropertyGroup/Version') | Select-Object -ExpandProperty '#text'
+            $xmlns = New-Object System.Xml.XmlNamespaceManager($csprojXml.NameTable)
+            $xmlns.AddNamespace('ns', $csprojXml.DocumentElement.NamespaceURI)
```
splatteredbits: What happens if there isn't a namespace? If the `xmlns` attribute on the project is missing? Add a test for that and see what happens.

jeffprice6: The existing test immediately above this one already handles this condition.

Issue #149, opened 2018-03-06
Publishing to nuget.org fails


splatteredbits:
Publishing to nuget.org fails when the NuGetPush task tries to verify if the package exists. It looks like nuget.org can take an hour or so for a package to show up where the task is checking.

> C:\projects\libgit2-powershell\whiskey.yml: BuildTasks[0]: NuGetPush: Failed to publish NuGet package LibGit2.PowerShell 0.9.0 to https://nuget.org/api/v2/package/package/LibGit2.PowerShell/0.9.0. When we checked if that package existed, we got a NotFound HTTP status code. Please see build output for more information.

PR #155, opened 2018-03-06
Fix issue #149: publishing to nuget.org fails.

splatteredbits:


Whiskey/Whiskey.psd1
```diff
@@ -149,6 +149,7 @@
             # ReleaseNotes of this module
             ReleaseNotes = @'
 * Fixed: PowerShell task fails when Whiskey is stored in a directory that isn't named "Whiskey".
+* Fixed: NuGetPush task fails when pushing to nuget.org.
```
jlariono: Can you point out how/where the NuGetPush task was fixed in regards to pushing to nuget.org? All I see is a property was added to allow skipping the check for the package after it's been uploaded.

splatteredbits: @jlariono Fixed. Ready for re-review.

PR #161, opened 2018-04-20
Parallel Task

splatteredbits:


Whiskey/Tasks/Invoke-WhiskeyParallelTask.ps1
```diff
@@ -0,0 +1,118 @@
...
+            $job | Add-Member -MemberType NoteProperty -Name 'TaskIndex' -Value $taskIdx
+            [void]$jobs.Add($job)
+    }
+
+    try
+    {
+        foreach( $job in $jobs )
+        {
+            Write-WhiskeyVerbose -Context $TaskContext -Message ('[{0}][{1}]  Waiting for background task.' -f $job.TaskIndex,$job.Name)
+            $job | Wait-Job | Receive-Job
```
jlariono: This line is synchronous right? Doesn't this mean we're only returning the job output in-order, as we're waiting for each task to complete in the order they were started? Meaning job 4's output won't be returned until we've returned job 1, 2, and 3, even though job 4 might have finished first? The documentation states that we're returning the output **"as each task finishes"** and not after all the tasks have finished in-order.
Also, what happens if job 2 finishes but job 1 fails? Do we get the output of job 2 before we stop Whiskey due to job 1's failure or is job 2's output simply lost when we fail the build?

splatteredbits: It means we wait until the first task finishes before checking if any other task finishes.

I've updated the task to loop through each job, waiting for it to finish. 

If a task fails, any task that is still running is stopped/aborted/cancelled and no output from them is returned. This is by design so that someone can go to the end of the built output to see what caused the job to fail.

I've also updated the task to support running multiple synchronous tasks in one background job.

splatteredbits:
@jlariono Ready for re-review.

Whiskey/Tasks/Invoke-WhiskeyParallelTask.ps1
```diff
@@ -0,0 +1,176 @@
...
+                    Sync-ObjectProperty -Source $originalContext -Destination $context -ExcludeProperty @( 'BuildMetadata', 'Version' )
+
+                    $context.BuildMetadata = $buildInfo
+                    $context.Version = $buildVersion
+
+                    $tasks = $using:queue['Tasks']
+                    foreach( $task in $tasks )
+                    {
+                        $taskName = $task.Keys | Select-Object -First 1
+                        $taskParameter = $task[$taskName]
```
jlariono: Will these two lines (123 & 124) write an error if someone uses a task without properties like `- DotNetPack` or `- NodeNspCheck`?

Whiskey/Tasks/Invoke-WhiskeyParallelTask.ps1
```diff
@@ -0,0 +1,176 @@
...
+                    Add-Member -MemberType NoteProperty -Name 'Completed' -Value $false
+                [void]$jobs.Add($job)
+        }
+
+        $lastNotice = (Get-Date).AddSeconds(-61)
+        while( $jobs | Where-Object { -not $_.Completed } )
+        {
+            foreach( $job in $jobs )
+            {
+                if( $job.Completed )
```
jlariono: This seems like a redundant check since the `while` loop's condition is already doing `$jobs | Where-Object { -not $_.Completed }`.

Whiskey/Tasks/Invoke-WhiskeyParallelTask.ps1
```diff
@@ -0,0 +1,176 @@
...
+                    if( $job.JobStateInfo.State -eq [Management.Automation.JobState]::Failed )
+                    {
+                        Stop-WhiskeyTask -TaskContext $TaskContext -Message ('Queue[{0}] failed. See previous output for error information.' -f $job.Name)
+                    }
+                }
+            }
+
+            if( $notified )
+            {
+                $notified = $true
```
jlariono: This should be set to `$false` not `$true`.

Whiskey/Tasks/Invoke-WhiskeyParallelTask.ps1
splatteredbits: The `while` loop is checking if all the jobs are completed. The `if( $job.Completed )` clause is checking if each individual job as completed since the `foreach` loops always loops through each job.

jlariono: Oh yes, in my head I was reading `while` as `Foreach-Object` for some reason.

Whiskey/en-US/about_Whiskey_Tasks.help.txt
```diff
@@ -13,6 +13,8 @@ LONG DESCRIPTION
 
     * `CopyFile`: copies files.
     * `Delete`: deletes files.
+    * `DotNetBuild`: runs `dotnet build` command.
+    * `DotNetPack`: runs `dotnet pack` command.
```
jlariono: Commit message says you wrote a test to ensure all tasks are included in this file but I don't see the test, just these additions to `about_Whiskey_Tasks.help.txt`.

splatteredbits: Forgot to add it. Fixed.

splatteredbits:
@jlariono Ready for next review.

splatteredbits:
@jlariono One final change to get tests to pass on the build server. Final review.

PR #174, opened 2018-05-30
Adding support for replacement tokens in .nuspec files

splatteredbits:
I need this functionality to get the Carbon module's build process migrated to Whiskey.

* Added "Properties" property to "NuGetPack" task so that tokens inside .nuspec files can be replaced. The "Properties" property should be a name/value mapping. Each name/value is passed to nuget.exe pack command's "-Properties" parameter.
* Added "PackageID" property to "NuGetPack" task to handle situations where a package's ID doesn't match the source .nuspec/.csproj file.
* Added "PackageVersion" property to "NuGetPack" task to allow customizing the package's version number.

### DIFF
--- a/Whiskey/Functions/Install-WhiskeyTool.ps1
+++ b/Whiskey/Functions/Install-WhiskeyTool.ps1
@@ -106,7 +106,7 @@ function Install-WhiskeyTool
 
         $waitedFor = (Get-Date) - $startedWaitingAt
         Write-Debug -Message ('[{0:yyyy-MM-dd HH:mm:ss}]  Process "{1}" obtained mutex "{2}" in {3}.' -f (Get-Date),$PID,$mutexName,$waitedFor)
-        $DebugPreference = 'SilentlyContinue'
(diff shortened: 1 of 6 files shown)


Whiskey/Functions/Install-WhiskeyTool.ps1
```diff
@@ -89,7 +89,7 @@ function Install-WhiskeyTool
     $startedWaitingAt = Get-Date
     $startedUsingAt = Get-Date
     $installLock = New-Object 'Threading.Mutex' $false,$mutexName
-    $DebugPreference = 'Continue'
+    #$DebugPreference = 'Continue'
```
jlariono: There is another uncommented `$DebugPreference = 'Continue'` at the bottom of this file in the `finally` block.

splatteredbits: Fixed.

splatteredbits:
@jlariono Ready for re-review.

PR #175, opened 2018-05-30
Fixed: ProGetUniversalPackage task should remove the temporary roboco…

jlariono:
…py output file after it's no longer necessary so it doesn't inadvertently get added to package with a "*.txt" wildcard Include item.

### DIFF
--- a/Whiskey/Tasks/New-WhiskeyProGetUniversalPackage.ps1
+++ b/Whiskey/Tasks/New-WhiskeyProGetUniversalPackage.ps1
@@ -84,24 +84,24 @@ function New-WhiskeyProGetUniversalPackage
         $parentPathParam['ParentPath'] = $sourceRoot
     }
 
-    if( -not $manifestProperties.ContainsKey('Name') )
+    if( -not $manifestProperties.ContainsKey('name') )
     {
-        $manifestProperties['Name'] = $name
+        $manifestProperties['name'] = $name
     }
 
-    if( -not $manifestProperties.ContainsKey('Title') )
+    if( -not $manifestProperties.ContainsKey('title') )
     {
-        $manifestProperties['Title'] = $name
+        $manifestProperties['title'] = $name
     }
 
-    if( -not $manifestProperties.ContainsKey('Description') )
+    if( -not $manifestProperties.ContainsKey('description') )
     {
-        $manifestProperties['Description'] = $TaskParameter['Description']
+        $manifestProperties['description'] = $TaskParameter['Description']
     }
 
-    if( -not $manifestProperties.ContainsKey('Version') )
+    if( -not $manifestProperties.ContainsKey('version') )
     {
-        $manifestProperties['Version'] = $version.SemVer2NoBuildMetadata.ToString()
+        $manifestProperties['version'] = $version.SemVer2NoBuildMetadata.ToString()
     }
 
     $tempRoot = $TaskContext.Temp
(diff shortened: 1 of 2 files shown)


Whiskey/Whiskey.psd1
```diff
@@ -150,6 +150,8 @@
 # 0.35.0
 
 * Created "SetVariableFromPowerShellDataFile" task for creating variables from values in PowerShell data files (e.g. .psd1 files, module manifests, etc.).
+* Fixed: `ProGetUniversalPackage` task should remove the temporary robocopy output file after it's no longer necessary so it doesn't inadvertently get added to package with a "*.txt" wildcard Include item.
```
splatteredbits: If somebody is including files under .output in their package, we should let them. You're putting the output file in your tasks's temp directory, which should get cleaned up automatically when the task completes. I'm curious how/where you saw this file in a package?

jlariono: My wording on the issue/fix was actually incorrect. What's actually happening is we use the root of the task temp directory as the temp staging area for the package contents. At the end of the task we zip up the entire task temp directory to create the upack package. So since the robocopy output files are being created in the task's temp directory, they're being zipped up with the package next to the upack.json.

That being said, I think the better solution is to move the temp staging area for the package one level deeper in the task's temp directory. Thoughts?

splatteredbits: Either approach works for me. Your decision.

jlariono: I moved the temp package staging directory one level below the root task temp directory to prevent any sort of misc task files from getting package in the future.

Whiskey/Whiskey.psd1
```diff
@@ -150,7 +150,7 @@
 # 0.35.0
 
 * Created "SetVariableFromPowerShellDataFile" task for creating variables from values in PowerShell data files (e.g. .psd1 files, module manifests, etc.).
-* Fixed: `ProGetUniversalPackage` task should remove the temporary robocopy output file after it's no longer necessary so it doesn't inadvertently get added to package with a "*.txt" wildcard Include item.
+* Fixed: `ProGetUniversalPackage` task now uses a separate package staging directory inside the root task temp directory to prevent any sort of miscellaneous task files from being included in the generated package.
```
splatteredbits: I would rephrase this as "`ProGetUniversalPackage` task no longer includes robocopy output text file in your package."

Whiskey/Tasks/New-WhiskeyProGetUniversalPackage.ps1
```diff
@@ -84,27 +84,29 @@ function New-WhiskeyProGetUniversalPackage
...
 
-    if( -not $manifestProperties.ContainsKey('Version') )
+    if( -not $manifestProperties.ContainsKey('version') )
     {
-        $manifestProperties['Version'] = $version.SemVer2NoBuildMetadata.ToString()
+        $manifestProperties['version'] = $version.SemVer2NoBuildMetadata.ToString()
     }
 
-    $tempRoot = $TaskContext.Temp
+    $tempRoot = Join-Path -Path $TaskContext.Temp -ChildPath ('Temp.{0}.upack' -f $name)
```
splatteredbits: I'm worried about paths that are too long. Maybe call the directory "upack"? Temp directory's are just for each task so there wouldn't be any collisions.

jlariono: Oh good point, fixed.

jlariono:
@splatteredbits Requested changes made.

PR #181, opened 2018-06-12
Feature/dotnet publish task

jlariono:
Created `DotNetPublish` task for building a project and its dependencies for deployment.


Whiskey/Tasks/Invoke-WhiskeyDotNetPublish.ps1
```diff
@@ -0,0 +1,81 @@
...
+
+    $verbosity = $TaskParameter['Verbosity']
+    if (-not $verbosity -and $TaskContext.ByBuildServer)
+    {
+        $verbosity = 'detailed'
+    }
+
+
+    $outputDirectory = $TaskParameter['OutputDirectory']
+    if ($outputDirectory -and ([IO.Path]::IsPathRooted($outputDirectory)))
```
splatteredbits: What happens if you pass an absolute path to this? Does it handle it? Can we just pass it as-is? Do we need to validate?

Whiskey/en-US/about_Whiskey_DotNetPublish_Task.help.txt
```diff
@@ -0,0 +1,57 @@
+TOPIC
+    about_Whiskey_DotNetPublish_Task
+
+SUMMARY
+    Builds and publishes a .NET Core project and all its dependencies for deployment.
+
+DESCRIPTION
+    The `DotNetPublish` tasks runs the `dotnet publish` command which compiles the application, reads through its dependencies specified in the project file, and publishes the resulting set of files to a directory. The `DotNetPublish` task's output is ready for deployment to a hosting system. Pass a list of solutions files or .NET Core project files to the `Path` property. If no files are provided to `Path`, then the .NET Core SDK will search for any solution or project files in the working directory and build those. If the `dotnet publish` command returns a non-zero exit code (i.e. compilation error), the build will fail.
```
splatteredbits: Replace the first two sentence with "The `DotNetPublish` task runs the `dotnet publish` command which packages "the application and its dependencies into a folder for deployment". Your application won't run until you've published it. The `dotnet build` command doesn't create a runnable bin directory. Pass a list of solution..."

Whiskey/Tasks/Invoke-WhiskeyDotNetBuild.ps1
```diff
@@ -32,9 +32,10 @@ function Invoke-WhiskeyDotNetBuild
     }
 
     $outputDirectory = $TaskParameter['OutputDirectory']
-    if ($outputDirectory)
+    if ($outputDirectory -and ([IO.Path]::IsPathRooted($outputDirectory)))
```
splatteredbits: What happens if you pass an absolute path to this? Does it handle it? Can we just pass it as-is? Do we need to validate?

Whiskey/Tasks/Invoke-WhiskeyDotNetPublish.ps1
jlariono: Absolute path works fine. I was just following our design pattern where we generally don't allow absolute paths in Whiskey task properties.

Whiskey/Tasks/Invoke-WhiskeyDotNetBuild.ps1
jlariono: Absolute path works fine. I was just following our design pattern where we generally don't allow absolute paths in Whiskey task properties.

Whiskey/Tasks/Invoke-WhiskeyDotNetPublish.ps1
splatteredbits: Since we're passing this to `dotnet.exe`, I don't think we should do any validation. Let `dotnet.exe` worry about validating it.

jlariono:
@splatteredbits Completed requested changes and made additional improvements to the `DotNet` tasks. Ready for another review.

Whiskey/Functions/Invoke-WhiskeyDotNetCommand.ps1
```diff
@@ -0,0 +1,92 @@
...
+    Use-CallerPreference -Cmdlet $PSCmdlet -SessionState $ExecutionContext.SessionState
+
+    $dotNetExe = $DotNetPath | Resolve-Path -ErrorAction 'Ignore'
+    if (-not $dotNetExe)
+    {
+        Write-Error -Message ('"{0}" does not exist.' -f $DotNetPath)
+        return
+    }
+
+    $loggerArgs = & {
```
splatteredbits: These are MSBuild parameters and won't work on other dotnet commands. These should get moved up into the `DotNetBuild` task.

Whiskey/Whiskey.psd1
```diff
@@ -157,8 +157,11 @@
 * Fixed: `ProGetUniversalPackage` task was creating upack.json property keys with uppercase letters when they are required to be lowercase.
 * Created `GitHubRelease` task for creating a release in GitHub. The task supports uploading files into the release.
 * Created `DotNetTest` task for running `dotnet test` on .NET Core test projects.
+* MSBuild task no longer includes the build configuration in the build log file.
```
splatteredbits: Change "build log file." to "build log file name.".

Whiskey/en-US/about_Whiskey_DotNetPublish_Task.help.txt
```diff
@@ -5,13 +5,15 @@ SUMMARY
...
 
     The task will automatically use the following options with the `dotnet publish` command and are not modifiable:
 
         * `--configuration=(Debug|Release)`: `Debug` when build run by a developer and `Release` when build run by a build server.
         * `-p:Version=$(WHISKEY_VERSION_SEMVER1)`: the semver version 1 version for the current build. Use a `Version` task in your whiskey.yml file to set the version number.
 
+    The task writes a detailed build log to the build output directory. If the `Path` property has no value, the file is named `dotnet.publish.log`. Otherwise, there will be a log file for each path named `dotnet.publish.FILE_NAME.log`, where `FILE_NAME` is the name of the file built.
```
splatteredbits: Remove. Only the `DotNetBuild` task should log to the output directory.

Whiskey/en-US/about_Whiskey_DotNetTest_Task.help.txt
```diff
@@ -13,6 +13,8 @@ DESCRIPTION
         * `--no-build`: Skip building the project prior to testing. Use Whiskey's `DotNetBuild` task to build the test project.
         * `--results-directory=$(WHISKEY_OUTPUT_DIRECTORY)`: By default, test results are not saved to a file. Use the `Logger` property to specify the logger URI/friendly name (e.g. `trx`). The results will be saved to a file in Whiskey's ".output" directory.
 
+    The task writes a detailed build log to the build output directory. If the `Path` property has no value, the file is named `dotnet.test.log`. Otherwise, there will be a log file for each path named `dotnet.test.FILE_NAME.log`, where `FILE_NAME` is the name of the file built.
```
splatteredbits: Remove. Only the `DotNetBuild` task should log to the output directory.

Whiskey/Functions/Invoke-WhiskeyDotNetCommand.ps1
```diff
@@ -0,0 +1,92 @@
+
+function Invoke-WhiskeyDotNetCommand
```
splatteredbits: I'm wondering if this should be a `DotNet` task that the other tasks call? It would only have two properties: `Command`, the name of the .NET command to run, and `Argument`, a list of arguments to pass to the command? Or maybe just the `ArgumentList` property?

```
Build:
- DotNet:
    ArgumentList: pack
```

What do you think?

Whiskey/Functions/Invoke-WhiskeyDotNetCommand.ps1
jlariono: I initially only had them for Build and Publish but then I tested these logging arguments with all the `DotNet` tasks and they worked, providing some relevant logging related to the dotnet command that was run. What it looked like was commands like `dotnet test` and `dotnet pack` were actually running MSBuild just with specific targets and so the logging args worked to create a log file.

I think [this documentation](https://docs.microsoft.com/en-us/dotnet/core/tools/cli-msbuild-architecture) confirms my suspicion of MSBuild running for all the `dotnet` commands.
>For example, the next version of Visual Studio will not call into dotnet restore (see note) command to restore dependencies for .NET Core projects, it will use the "Restore" target directly. Since these are MSBuild targets, you can also use raw MSBuild to execute them using the dotnet msbuild command.

With that in mind, should I keep the logging args for all `dotnet` commands or use them only for `DotNetBuild` and `DotNetPublish`?

Whiskey/Functions/Invoke-WhiskeyDotNetCommand.ps1
jlariono: I do like that idea, it's clean and adds additional flexibility to the whiskey.yml but at the same time it adds additional overhead of setting up and running a whole other task. But that additional overhead might be negligible in the grand scheme of things.

Though it does add the possibility of running into random edge cases where a task is calling another task like we had when we added the `Parallel` task.

Whiskey/Functions/Invoke-WhiskeyDotNetCommand.ps1
splatteredbits: Leave them. I assumed `dotnet.exe` only called MSBuild when building.

Whiskey/Functions/Invoke-WhiskeyDotNetCommand.ps1
splatteredbits: Let's just be done. Leave as-is.

splatteredbits:
@jlariono Approve, but the build is failing. You'll want to fix that first. I'll probably need to re-approve.

jlariono:
@splatteredbits All fixed up.

jlariono:
@splatteredbits Had to fix the merge conflict due to the PR merge today. Need one more approval. <sup>sorry<sup/>

splatteredbits:
@jlariono Zapproved!

PR #187, opened 2018-06-26
Allow ProGet to exclude files/paths

splatteredbits:


Whiskey/Whiskey.psd1
```diff
@@ -12,7 +12,7 @@
     RootModule = 'Whiskey.psm1'
 
     # Version number of this module.
-    ModuleVersion = '0.35.0'
+    ModuleVersion = '0.36.0'
```
jlariono: I think the version needs to be 0.37.0 since I inadvertently published a 0.36.0 version prematurely.

Test/PublishProGetUniversalPackage.Tests.ps1
```diff
@@ -355,6 +373,19 @@ Describe 'Publish-WhiskeyProGetUniversalPackage.when user specifies files to pub
     ThenPackageNotPublished 'myfile2.upack'
 }
 
+Describe 'Publish-WhiskeyProGetUniversalPackage.when user specifies files to publish and excluding some' {
```
jlariono: Should we add a warning to the task if the exclude wildcard pattern ends up matching all files and nothing gets published?

Test/PublishProGetUniversalPackage.Tests.ps1
```diff
@@ -342,6 +348,18 @@ Describe 'Publish-WhiskeyProGetUniversalPackage.when user publishes default file
     ThenPackagePublishedWithDefaultTimeout
 }
 
+Describe 'Publish-WhiskeyProGetUniversalPackage.when user excludes files' {
```
jlariono: Probably should rename the Describe blocks now that you renamed the script.

PR #190, opened 2018-07-17
Bugfix/msbuild version multiple installed

jlariono:
Fixed: `MSBuild` task fails when given `Version` property and multiple installs of that version of MSBuild exist on the system.

### DIFF
--- a/Whiskey/Tasks/Invoke-WhiskeyMSBuild.ps1
+++ b/Whiskey/Tasks/Invoke-WhiskeyMSBuild.ps1
@@ -6,39 +6,39 @@ function Invoke-WhiskeyMSBuild
         [Whiskey.Context]
         # The context this task is operating in. Use `New-WhiskeyContext` to create context objects.
         $TaskContext,
-        
+
         [hashtable]
         # The parameters/configuration to use to run the task. Should be a hashtable that contains the following item(s):
-        # 
+        #
         # * `Path` (Mandatory): the relative paths to the files/directories to include in the build. Paths should be relative to the whiskey.yml file they were taken from.
         $TaskParameter
     )
 
-    Set-StrictMode -version 'Latest'  
+    Set-StrictMode -version 'Latest'
     Use-CallerPreference -Cmdlet $PSCmdlet -SessionState $ExecutionContext.SessionState
 
     #setup
     $nuGetPath = Install-WhiskeyNuGet -DownloadRoot $TaskContext.BuildRoot -Version $TaskParameter['NuGetVersion']
-    
+
     # Make sure the Taskpath contains a Path parameter.
     if( -not ($TaskParameter.ContainsKey('Path')) -or -not $TaskParameter['Path'] )
     {
-        Stop-WhiskeyTask -TaskContext $TaskContext -Message ('Element ''Path'' is mandatory. It should be one or more paths, relative to your whiskey.yml file, to build with MSBuild.exe, e.g. 
-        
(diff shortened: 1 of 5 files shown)


Whiskey/Tasks/Invoke-WhiskeyMSBuild.ps1
```diff
@@ -6,39 +6,39 @@ function Invoke-WhiskeyMSBuild
...
-    
+
     $path = $TaskParameter['Path'] | Resolve-WhiskeyTaskPath -TaskContext $TaskContext -PropertyName 'Path'
 
     $msbuildInfos = Get-MSBuild | Sort-Object -Descending 'Version'
     $version = $TaskParameter['Version']
     if( $version )
     {
-        $msbuildInfo = $msbuildInfos | Where-Object { $_.Name -eq $version }
+        $msbuildInfo = $msbuildInfos | Where-Object { $_.Name -eq $version } | Select-Object -First 1
```
jeffprice6: If a version's "Build Tools" and "Professional" are both installed, do they always return in a certain order? Does it matter what order they are installed or which one is selected to use? I'm not sure, but am wondering if it would be necessary for a secondary sort on line 37 to enforce the desired order.

build.ps1
```diff
@@ -61,7 +61,7 @@ Import-Module (Join-Path -Path $PSScriptRoot -ChildPath 'Whiskey\VSSetup')
 . (Join-Path -Path $PSScriptRoot -ChildPath 'Whiskey\Functions\Use-CallerPreference.ps1')
 . (Join-Path -Path $PSScriptRoot -ChildPath 'Whiskey\Functions\Get-MSBuild.ps1')
 
-$msbuild = Get-MSBuild -ErrorAction Ignore | Where-Object { $_.Name -eq '15.0' }
+$msbuild = Get-MSBuild -ErrorAction Ignore | Where-Object { $_.Name -eq '15.0' } | Select-Object -First 1
```
jeffprice6: Same here (if the order does matter).

Whiskey/Tasks/Invoke-WhiskeyMSBuild.ps1
jlariono: I decided against sorting them into a specific order because we have no idea of know which of the "Professional" (bundled with Visual Studio) or "BuildTools" (standalone) MSBuild install is preferred.

A developer might have setup all his build tools manually with the standalone install and then enabled additional required build tools components for the standalone install. That means the build tools installed by Visual Studio exist but don't have all the necessary components enabled to run a build successfully. Or the reverse of that could be true. So we have no way of knowing which is the correct install of the build tools to use.

So with the lack of a good reason to choose or default one of the build tools install paths over the other, I opted to just let it return the first one of the desired version regardless of the order in which it resolved the paths.

If the build tools that were selected during the build are missing required components then the user can either add the missing components or uninstall that secondary instance of the build tools so that the correct one is used during the build.

jeffprice6: Sounds good. I thought that might be the case, just wasn't sure if there was an always correct priority..

jlariono: I'm still not convinced that it's the best solution though. I'll wait to merge to get @splatteredbits thoughts.

jlariono: Per our conversations offline, we'll keep the current implementation as-is.

PR #201, opened 2018-11-08
`Install-WhiskeyPowerShellModule` function initializes NuGet package …

jeffprice6:
…provider

### DIFF
--- a/Whiskey/Functions/Resolve-WhiskeyPowerShellModule.ps1
+++ b/Whiskey/Functions/Resolve-WhiskeyPowerShellModule.ps1
@@ -3,14 +3,14 @@ function Resolve-WhiskeyPowerShellModule
     <#
     .SYNOPSIS
     Searches for a PowerShell module using PowerShellGet to ensure it exists and returns the resulting object from PowerShellGet.
-    
+
     .DESCRIPTION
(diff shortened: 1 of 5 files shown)


splatteredbits:
@jeffprice6 This needs to happen everywhere we use the PackageManagement and PowerShellGet before we use any cmdlet from those modules. We need this bootstrap in `Resolve-WhiskeyPowerShellModule`, too.

Whiskey/Whiskey.psd1
```diff
@@ -144,7 +144,7 @@
 
             # ReleaseNotes of this module
             ReleaseNotes = @'
-* Fixed: `Install-WhiskeyPowerShellModule` function fails when NuGet package provider is not initialized.
+* Fixed: `Install-WhiskeyPowerShellModule` and `Resolve-WhiskeyPowerShellModule` functions may when NuGet package provider is not initialized.
```
splatteredbits: Fix grammar: "may *fail* when".

Whiskey/Functions/Resolve-WhiskeyPowerShellModule.ps1
```diff
@@ -36,6 +36,10 @@ function Resolve-WhiskeyPowerShellModule
     Set-StrictMode -Version 'Latest'
     Use-CallerPreference -Cmdlet $PSCmdlet -SessionState $ExecutionContext.SessionState
 
+    Import-WhiskeyPowerShellModule -Name 'PackageManagement', 'PowerShellGet'
```
splatteredbits: Yikes. That's important, otherwise the system-installed modules are auto-loaded! Good catch.

PR #211, opened 2018-12-28
Removing Usages of 7-Zip

splatteredbits:
Removed most usages of 7-Zip. We still need it to install Node.js on Windows, since the Node package contains paths that are too long for Windows. But, if we need it, we download 7-Zip from nuget.org.


Whiskey/Tasks/New-WhiskeyZipArchive.ps1
```diff
@@ -0,0 +1,201 @@
...
+        $buildRootRegex = '^{0}(\\|/)' -f [regex]::Escape($TaskContext.BuildRoot)
+        if( $archivePath -notmatch $buildRootRegex )
+        {
+            Stop-WhiskeyTask -TaskContext $TaskContext -Message ('ArchivePath: path to ZIP archive "{0}" is outside the build root directory. Please change this path so it is under the "{1}" directory. We recommend using a relative path, as that will always be resolved relative to the build root directory.' -f $archivePath,$TaskContext.BuildRoot)
+            return
+        }
+    }
+    else
+    {
+        $archivePath = Join-Path -Path $TaskContext.BuildRoot -ChildPath $archivePath
```
DecoyJoe: Do we not care to validate that the relative path is inside the build root? E.g. Someone specifies `..\..\File.zip`

Whiskey/Tasks/New-WhiskeyZipArchive.ps1
```diff
@@ -0,0 +1,201 @@
...
+            {
+                $destinationItemName = $item[$key]
+                $sourcePath = $key
+            }
+        }
+        else
+        {
+            $sourcePath = $item
+        }
+        $pathparam = 'Path'
```
DecoyJoe: Unused variable

Test/Zip.Tests.ps1
```diff
@@ -0,0 +1,571 @@
...
+    Remove-Module -Force -Name Zip -ErrorAction Ignore
+}
+
+function Install-Zip
+{
+    param(
+        $BuildRoot
+    )
+
+    # Copy ProGetAutomation in place otherwise every test downloads it from the gallery
```
DecoyJoe: Copied comment references `ProGetAutomation` in `Zip` task test file.

Test/Zip.Tests.ps1
```diff
@@ -0,0 +1,571 @@
...
+    Path:
+    - "*.ps1"
+'@
+    ThenTaskFails 'is an invalid compression level'
+}
+
+Describe 'Zip.when package has empty directories' {
+    Init
+    GivenARepositoryWithItems 'root.ps1','dir1\one.ps1','dir1\emptyDir2\text.txt'
+    GivenARepositoryWithItems 'dir1\emptyDir1' -ItemType 'Directory'
```
DecoyJoe: Test case is not accurate as described because the function `GivenARepositoryWithItems` never implements the `ItemType` parameter, resulting in `dir1\emptyDir1` being created as a File rather than a Directory.

Test/Zip.Tests.ps1
```diff
@@ -0,0 +1,571 @@
...
+
+    It ('should fail with error message that matches "{0}"' -f $error) {
+        $Global:Error | Should match $error
+    }
+}
+
+function ThenTaskSucceeds
+{
+    It ('should not throw an error message') {
+        $Global:Error | Should BeNullOrEmpty
```
DecoyJoe: Mixing Pester 3 and 4 assertion syntax style in this file. `BeNullOrEmpty` should be `-BeNullOrEmpty`

Test/Zip.Tests.ps1
```diff
@@ -0,0 +1,571 @@
...
+    
+    $Global:Error.Clear()
+
+    try
+    {
+        Invoke-WhiskeyTask -TaskContext $taskContext -Parameter $taskParameter -Name 'Zip'
+    }
+    catch
+    {
+        $threwException = $true
```
DecoyJoe: Sets `$threwException` in the local function scope, instead of `$script:` scope because it was defined earlier on Line 175.

Also, this variable is never checked in a test assertion in this file.

Test/Zip.Tests.ps1
```diff
@@ -0,0 +1,571 @@
...
+    WhenPackaging @'
+Build:
+- Zip:
+    ArchivePath: Zip.zip
+    Path:
+    - .
+    Include:
+    - "*.fubar"
+'@
+    ThenArchiveShouldInclude
```
DecoyJoe: This does not assert that the archive is empty, instead with no given value it just doesn't check anything.

Whiskey/Tasks/New-WhiskeyProGetUniversalPackage.ps1
```diff
@@ -172,99 +171,117 @@ function New-WhiskeyProGetUniversalPackage
...
     {
-        Copy-ToPackage -Path $TaskParameter['ThirdPartyPath'] -AsThirdPartyItem
+        $manifestProperties['title'] = $TaskParameter['Name']
     }
 
-    $badChars = [IO.Path]::GetInvalidFileNameChars() | ForEach-Object { [regex]::Escape($_) }
-    $fixRegex = '[{0}]' -f ($badChars -join '')
-    $fileName = '{0}.{1}.upack' -f $name,($version.SemVer2NoBuildMetadata -replace $fixRegex,'-')
+    $outFileDisplay = $outFile -replace ('^{0}' -f [regex]::Escape($TaskContext.BuildRoot)),''
+    $outFileDisplay = $outFileDisplay.Trim('\','/')
```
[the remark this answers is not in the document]
splatteredbits: To support whatever directory separator chars people use. I tried to use `DirectorySeparatorChar` and `AltDirectorySeparatorChar`, but on Linux, both of those are `/`.

Whiskey/Tasks/New-WhiskeyZipArchive.ps1
```diff
@@ -0,0 +1,201 @@
+
+function New-WhiskeyZipArchive
```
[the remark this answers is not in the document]
splatteredbits: I've been thinking about this. Not quite sure I'm ready to give up the implementation name to file name map. Regardless, this should be a mass PR where we update all the task files and test files.

Whiskey/Tasks/New-WhiskeyZipArchive.ps1
splatteredbits: Added test case.

splatteredbits:
@DecoyJoe Changes made. Ready for next review.

Whiskey/Functions/Install-WhiskeyTool.ps1
```diff
@@ -253,8 +253,22 @@ function Install-WhiskeyTool
 
                             if( -not (Test-Path -Path $nodePath -PathType Leaf) )
                             {
-                                Write-Verbose -Message ('{0} x {1} -o{2} -y' -f $7z,$nodeZipFile,$nodeRoot)
-                                & $7z 'x' $nodeZipFile ('-o{0}' -f $nodeRoot) '-y'
+                                #if( $IsWindows )
```
DecoyJoe: Sadly you'll need to un-comment this code because it caused the build to fail, I guess we'll need to rely on 7-Zip for this.

PR #214, opened 2019-01-15
Download Package Management Modules

splatteredbits:
Download the package management modules at build time. They are really big and probably shouldn't be checked into source.


build.ps1
```diff
@@ -102,6 +102,29 @@ foreach( $assembly in (Get-ChildItem -Path $whiskeyOutBinPath -Filter '*.dll') )
...
+# fixed, change version numbers to wildcards on the patch version.
+$nestedModules = @{
+                        'PackageManagement' = '1.2.4';
+                        'PowerShellGet' = '2.0.3';
+                 }
+$whiskeyRoot = Join-Path -Path $PSScriptRoot -ChildPath 'Whiskey'
+foreach( $nestedModuleName in $nestedModules.Keys )
+{
+    $moduleRoot = Join-Path -Path $whiskeyRoot -ChildPath $nestedModuleName
+    if( -not (Test-Path -Path $moduleRoot -PathType Container) )
```
DecoyJoe: Should we rather check for the module's `.psd1` file? Otherwise an empty module directory would appear as the module already being installed.

PR #215, opened 2019-01-15
Install local .NET Core

splatteredbits:
We should install a local .NET Core version so that developers don't get build failures when they have the wrong global version of .NET Core installed.


build.ps1
```diff
@@ -40,11 +40,38 @@ if( Test-Path -Path ('env:APPVEYOR') )
...
     $buildInfo = '+{0}.{1}.{2}' -f $env:APPVEYOR_BUILD_NUMBER,$branch,$commitID
     $MSBuildConfiguration = 'Release'
+
+    $env:DOTNET_SKIP_FIRST_TIME_EXPERIENCE = 'true'
+}
+
+$minDotNetVersion = [version]'2.1.503'
+$dotnetVersion = $null
+$dotnetInstallDir = Join-Path -Path $PSScriptRoot -ChildPath '.dotnet'
+$dotnetPath = Join-Path -Path $dotnetInstallDir -ChildPath 'dotnet.exe'
```
DecoyJoe: Windows specific logic that we'll have to fix later when we run a build on Linux/MacOS.

splatteredbits: Yes. I'll be fixing/addressing it over in `feature/cross-platform`. I was trying to be fast on this change.

PR #229, opened 2019-01-30
Cross Platform Support

splatteredbits:


Test/Invoke-WhiskeyExec.Tests.ps1
```diff
@@ -48,7 +48,15 @@ function GivenADirectory
     New-Item -Path (Join-Path -Path (Get-BuildRoot) -ChildPath $DirectoryPath) -ItemType 'Directory' -Force | Out-Null
 }
 
-function GivenExecutableFile
+function GivenExecutable
```
DecoyJoe: This function doesn't do anything and isn't used in any tests.

Whiskey/Tasks/Invoke-WhiskeyExec.ps1
```diff
@@ -40,21 +40,28 @@ function Invoke-WhiskeyExec
...
+                    else 
+                    {
+                        Join-Path -Path (Get-Location).Path -ChildPath $path
+                        Join-Path -Path $TaskContext.BuildRoot -ChildPath $path
+                    }
+            } |
+            Where-Object { Test-Path -path $_ -PathType Leaf } |
+            Select-Object -First 1 |
+            Resolve-Path |
+            Select-Object -ExpandProperty 'ProviderPath'
```
DecoyJoe: We need to call `Select-Object -First 1` before this line because if the user specified a wildcard value for their `WorkingDirectory` property, or worse the path contains literal wildcard values, that resolves to multiple path items then `Resolve-Path` will return more than one object and the task will fail to execute the command.

Or we use the `-LiteralPath` parameter with `Resolve-Path`

Whiskey/Functions/Uninstall-WhiskeyPowerShellModule.ps1
```diff
@@ -50,9 +50,12 @@ function Uninstall-WhiskeyPowerShellModule
...
 
-    $psmodulesDirEmpty = $null -eq (Get-ChildItem -Path $modulesRoot -File -Recurse)
-    if( $psmodulesDirEmpty )
+    if( (Test-Path -Path $modulesRoot -PathType Container) )
     {
-        Remove-Item -Path $modulesRoot -Recurse -Force
+        $psmodulesDirEmpty = $null -eq (Get-ChildItem -Path $modulesRoot -File -Recurse)
+        if( $psmodulesDirEmpty )
+        {
+            Remove-Item -Path $modulesRoot -Recurse -Force
```
DecoyJoe: We should have better discipline about when we use `-Path` vs `-LiteralPath`, otherwise we might run into a case where we accidentally delete a directory/file we didn't intend because `-Path` resolves wildcard characters.

Test/Invoke-WhiskeyExec.Tests.ps1
splatteredbits: ‽

Whiskey/Tasks/Invoke-WhiskeyExec.ps1
splatteredbits: I added a check if `Resolve-Path` returns multiple paths. I can see users wanting to use wildcards to find an executable. We just want to make sure it resolves to *one* executable.

Whiskey/Functions/Uninstall-WhiskeyPowerShellModule.ps1
splatteredbits: In this case, we want to use `Path` because we are using a `*.*.*` wildcard to remove all versions of a module.

PR #231, opened 2019-02-06
Redirect all verbose messages to null when importing a module

DecoyJoe:
The "Parallel" task displays all verbose messages that were written
during the background job, regardless of the "$VerbosePreference" value.
Therefore when importing PowerShell modules we need to redirect all
verbose messages to null to prevent the import messages from entering
the verbose stream and being returned at the end of the Parallel task.


DecoyJoe:
Added another fix to this PR:

Fixed the "ProGetUniversalPackage" placing files in a "." directory when
using the `source_dir: .` syntax to specify the package root as the
destination path for items in a package.

DecoyJoe:
@splatteredbits Build fixed, I'll need another approval.

PR #244, opened 2019-05-31
PFORM-36: bump version to 0.42.0

pnstickne:
- Minor feature enchancement (adding NUnit2 support to NUnit3 task)

### DIFF
--- a/Whiskey/Whiskey.psd1
+++ b/Whiskey/Whiskey.psd1
@@ -12,7 +12,7 @@
     RootModule = 'Whiskey.psm1'
 
     # Version number of this module.
-    ModuleVersion = '0.41.1'
+    ModuleVersion = '0.42.0'
 
     # ID used to uniquely identify this module
     GUID = '93bd40f1-dee5-45f7-ba98-cb38b7f5b897'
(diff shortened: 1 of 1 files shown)


Whiskey/Whiskey.psd1
```diff
@@ -153,7 +153,20 @@
 
             # ReleaseNotes of this module
             ReleaseNotes = @'
+* NUnit3 task supports running NUnit2 tests and uses version 3.10 of ConsoleRunner.
```
DecoyJoe: These are the release notes from the previous `0.41.0` release. The release notes should simply have:
```
* Fixed: `Version` task uses the last matching item in the prerelease label branch map instead of the first match.
* <Insert your release notes here>
```

Whiskey/Whiskey.psd1
```diff
@@ -154,6 +154,7 @@
             # ReleaseNotes of this module
             ReleaseNotes = @'
 * Fixed: `Version` task uses the last matching item in the prerelease label branch map instead of the first match.
+* NUnit3 task supports running NUnit2 tests and uses version 3.10 of ConsoleRunner.
```
DecoyJoe: Add one more line for your other change:
```
* Added a `ResultFormat` property to the `NUnit3` task for customizing the result format of NUnit itself. See NUnit for supported formats. Defaults to 'nunit3'.
```

PR #246, opened 2019-06-11
Task Aliases

splatteredbits:
If I want to change a task's name, there should be a way to make the old name usable with the new name and warn users to use the task's new name.


Whiskey/Functions/Invoke-WhiskeyTask.ps1
```diff
@@ -115,6 +115,15 @@ function Invoke-WhiskeyTask
 
     $task = $knownTasks | Where-Object { $_.Name -eq $Name }
 
+    if( -not $task )
+    {
+        $task = $knownTasks | Where-Object { $_.Aliases -contains $Name }
```
DecoyJoe: Should we also check to make sure this resolved to a single task? What if there are 2 more tasks that accidentally share a common alias?

appveyor.yml
```diff
@@ -17,11 +17,14 @@ environment:
...
 build_script:
+#- sh: curl -sflL 'https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-ssh.sh' | bash -e -
 - pwsh: >-
     # $blockRdp = $true; iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
     
     # On Linux, progress bars still show when running on the build server and take *ages* to render and they screw up the displayed output.
-    $ProgressPreference = 'SilentlyContinue'
+    $Global:ProgressPreference = 'SilentlyContinue'
+    
+    Write-Verbose -Message ([Environment]::UserInteractive) -Verbose
```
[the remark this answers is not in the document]
splatteredbits: Removed the `Write-Verbose` statement.

Whiskey/Functions/Invoke-WhiskeyTask.ps1
splatteredbits: Fixed. I also added a test and check for duplicate task names.

Whiskey/Functions/Invoke-WhiskeyTask.ps1
```diff
@@ -118,6 +118,12 @@ function Invoke-WhiskeyTask
     if( -not $task )
     {
         $task = $knownTasks | Where-Object { $_.Aliases -contains $Name }
+        $taskCount = ($task | Measure-Object).Count
+        if( $taskCount -gt 1 )
+        {
+            Stop-WhiskeyTask -TaskContext $TaskContext -Message ('Found {0} tasks with alias "{1}". Please update to use one of these task names: {3}.' -f $taskCount,$Name,$TaskContext.ConfigurationPath.FullName,(($task | Select-Object -ExpandProperty 'Name') -join ', '))
```
DecoyJoe: Number of tokens doesn't match the number of string format parameters.

PR #247, opened 2019-06-17
Obsolete Tasks

splatteredbits:


Whiskey/en-US/about_Whiskey_Writing_Tasks.help.txt
```diff
@@ -208,6 +208,16 @@ LONG DESCRIPTION
 
         [Whiskey.Task('TASK_NAME',Aliases=('DeprecatedName'),WarnWhenUsingAlias=$true)]
 
+    # Deprecating Tasks
+
+    To deprecate a task, set the `Obsolete` property on its `Whiskey.Task` attribute:
+
+        [Whiskey.Task('TASK_NAME',Obsolete)]
+
+    When someone uses your task, Whiskey will write a warning that the task is obsolete and shouldn't be used. You can customize the message shown to users with the `ObsoleteMessage` property:
```
DecoyJoe: Nit: Doesn't `obsolete` imply that something is old/out-of-date because something newer/better is available? This may not always be the case when deprecating a task, a task may simply be retired without any replacement. Would `deprecated` be a better property name?

Whiskey/Tasks/Invoke-WhiskeyNodeNspCheck.ps1
```diff
@@ -1,9 +1,9 @@
 
 function Invoke-WhiskeyNodeNspCheck
 {
-    [Whiskey.Task("NodeNspCheck")]
-    [Whiskey.RequiresTool("Node", "NodePath", VersionParameterName='NodeVersion')]
-    [Whiskey.RequiresTool("NodeModule::nsp", "NspPath", VersionParameterName="Version")]
+    [Whiskey.Task('NodeNspCheck',Obsolete,ObsoleteMessage='The "NodeNspCheck" task is obsolete and will be removed in a future version of Whiskey. Please use the "NpmAudit" task instead. The NSP project died in September 2018 and was replaced with the `npm audit` command.')]
```
DecoyJoe: `NpmAudit` is not a valid task. Should be pointing to the `Npm` task.

Whiskey/en-US/about_Whiskey_Writing_Tasks.help.txt
splatteredbits: `Obsolete` is what .NET uses. There may *not* be a new/better way. Usually there is. But deprecated is too specific. Obsolete leaves it open as to why someone shouldn't use a task anymore.

Whiskey/Tasks/Invoke-WhiskeyNodeNspCheck.ps1
splatteredbits: Fixed.

PR #248, opened 2019-06-20
Standardize task test fixture file names; be intentional about releasing pre-release versions; test improvements on Linux

splatteredbits:


whiskey.yml
```diff
@@ -11,7 +11,7 @@ Build:
 - Version:
     Path: Whiskey\Whiskey.psd1
     Prerelease:
-    - develop: beta$(WHISKEY_BUILD_NUMBER)
+    - release: beta$(WHISKEY_BUILD_NUMBER)
```
DecoyJoe: While I know this is following the Gitflow model, it's confusing at first since the `release` branch sounds like where you ship final releases.

splatteredbits: What should we use for the branch name? `beta`?

DecoyJoe: `beta` works or even the aptly name `prerelease` ?

splatteredbits: Changed to `prerelease`.

splatteredbits:
> I vote we also change the task functions and filenames to match the task names. We always reference the tasks by their task name, not function name. In addition, the task functions are not exposed outside the module and thus don't necessarily need to follow approved Verb-Noun formats.

Done.

PR #258, opened 2019-08-14
Installing modules fails if there is a feed that requires credentials

splatteredbits:


appveyor.yml
```diff
@@ -14,14 +14,21 @@ environment:
   matrix:
   - APPVEYOR_BUILD_WORKER_IMAGE: Ubuntu1804
     LANG: us_US.UTF-8
-  - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+  - APPVEYOR_BUILD_WORKER_IMAGE: Previous Visual Studio 2017
```
DecoyJoe: It would be nice to have a comment here with a link to the GitHub issue that is causing us to have to use the `Previous` image.

Issue #252, opened 2019-07-02
PublishPowerShellModule task fails if a PSRepository is already registered with a different name


DecoyJoe:
If the `PublishPowerShellModule` task is given a `RepositoryUri` and the system already has a registered PSRepository for that Uri but with a different name, the task will fail.

Update the `PublishPowerShellModule` task so that when given a `RepositoryUri` property, it checks to see if any existing registered PSRepository points to that same uri. If so, then write a warning and use the existing registered repository instead of trying to re-register the repository with a different name (and thus failing the build).

PR #264, opened 2019-08-28
Issue #252: fixed bug when trying to register an already existing PSR…

ericharper1:
…epository under a different name.

### DIFF
--- a/Test/PublishPowerShellModule.Tests.ps1
+++ b/Test/PublishPowerShellModule.Tests.ps1
@@ -23,7 +23,7 @@ function GivenCredential
         [string]$WithID
     )
 
-    $script:credentials[$WithID] = $Credential
+    $credentials[$WithID] = $Credential
 }
 
 function GivenNoApiKey
--- a/Test/PublishPowerShellModule.Tests.ps1
+++ b/Test/PublishPowerShellModule.Tests.ps1
@@ -5,6 +5,7 @@ Set-StrictMode -Version 'Latest'
 $apikey = $null
 $apikeyID = $null
 $repositoryName = $null
+$repositoryUri = $null
 $prerelease = $null
 $context = $null
 $credentials = @{ }
@@ -67,10 +68,12 @@ function GivenRegisteringFails
 function GivenRepository
 {
     param(
-        $Named
+        $Named,
+        $Uri
     )
 
     $script:repositoryName = $Named
+    $script:repositoryUri = $Uri
 }
 
 function Initialize-Test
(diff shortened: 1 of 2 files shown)


Whiskey/Tasks/PublishPowerShellModule.ps1
```diff
@@ -76,23 +76,42 @@ function Publish-WhiskeyPowerShellModule
...
+                                      -ID $TaskParameter['CredentialID'] `
+                                      -PropertyName 'CredentialID'
         }
 
-        Register-PSRepository -Name $repositoryName -SourceLocation $publishLocation -PublishLocation $publishLocation -InstallationPolicy Trusted -PackageManagementProvider NuGet @credentialParam -ErrorAction Stop @commonParams
+        $exists = $registeredRepositories | Where-Object { $_.SourceLocation -eq $publishLocation }
+        if( $exists )
+        {
+            $repositoryName = $exists.Name 
+            Write-Warning -Message ('The uri "{0}" is already a registered repository under a different name. Please update your whiskey.yml file.' -f $publishLocation)
```
splatteredbits: Change `uri` to `URI`.

Use `Write-WhiskeyWarning` instead of `Write-Warning`.

Change "is already a registered repository under a different name" to `is registered as the "{1}" repository. Please update your whiskey.yml file to use this repository name.`.

Test/PublishPowerShellModule.Tests.ps1
```diff
@@ -177,15 +181,18 @@ function Invoke-Publish
...
-        )
-        #$DebugPreference = 'Continue'
-        Write-Debug -Message ('Name  expected  {0}' -f $repoName)
-        Write-Debug -Message ('      actual    {0}' -f $Name)
-        return ($Name -eq $repoName) 
+    $repoName = $script:repositoryName
+    $repoUri = $script:repositoryUri
+
+    Mock -CommandName 'Get-PSRepository' -ModuleName 'Whiskey' -MockWith {
+        $repositoryObject = New-Object System.Object
```
splatteredbits: Instead:

```
    return [pscustomobject]@{
        'Name' = $repoName;
        'SourceLocation' = $repoURi;
    }
```

Test/PublishPowerShellModule.Tests.ps1
```diff
@@ -17,15 +17,13 @@ function GivenCredential
...
+        [pscredential]$Credential,
 
         [Parameter(Mandatory)]
-        [string]
-        $WithID
+        [string]$WithID
     )
 
-    $credentials[$WithID] = $Credential
+    $script:credentials[$WithID] = $Credential
```
splatteredbits: Since you're setting the contents of the hashtable, you don't need `script:`.

You only need the `script:` scope when changing the value of a script-level variable. PowerShell sets variables locally even if there is an equivalent global variable.

PR #283, opened 2019-09-30
Move Whiskey's documentation to its GitHub wiki.

splatteredbits:
I've moved all of Whiskey's documentation to the wiki.


splatteredbits:
My philosophy is to make documentation version independent. If a new feature is added, note what version it was added in. If a feature was taken away, note what version it was removed (unless it is an obsolete/deprecated feature, then it's OK to just remove it).

I would love a generate documentation task. I think we should go with PlatyPS. You write docs in Markdown, and PlatyPS converts to PowerShell's XML-based MAML format. We could then have another task that convert it to GitHub wiki syntax. I prefer the GitHub wiki since it keeps everythint all together.

That being said, today, I think it is more valuable to put the docs in the wiki than in the PowerShell. I find myself referencing it all the time and viewing the docs in the plain text format on GitHub is painful. Plus, with docs in the wiki, there's no overhead to commmitting, which starts a build, which requires a PR. I can just... write the docs. It makes it much easier to write them.

DecoyJoe:
I still like the idea of being able to explore docs for each version, even if it's just in the repo (by using version tags). However version independent works too, though I do have concerns it might get messy/confusing over-time with **Added in:**/**Removed in:**/**Optional in:** all over the place?

Putting the docs in the wiki does indeed make them easier to view in most cases. However as a developer, it's helpful to have them in the repo as well as I often reference them directly in my editor as I work in Whiskey. There's even [Show-Markdown](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/show-markdown?view=powershell-6) in PowerShell 6 that allows you to render markdown directly in the console.

What's the new documentation workflow and review process? Can you write a documentation readme that explains where to find it and a guide for developers on the process to add/change documentation?

If/when we do create a task to generate documentation, is it going to be painful to extract the documentation out of the wiki and back into the repo?

splatteredbits:
The documentation on the wiki is written in Markdown, so it shouldn't be too painful. That's what PlatyPS does. It wants function/cmdlet documentation in a specific format, but I've kept all that in the repo. If someone's working with the API, they've got PowerShell open, Whiskey imported, and are doing work.

I've just moved the about topics to the wiki. If I'm just working on a whiskey.yml file, I just want to know what tasks are available and their syntax. I think the web is the best fit for that. We could always also export the wiki and ship it with the module.

splatteredbits:
The Added In/Removed In notations don't get too egregious. That's how I've been writing the Carbon docs and there aren't that many.

DecoyJoe:
Alright, given that, I'm amenable to these changes. I'll wait for a green build.

PR #290, opened 2019-10-15
Fixed: .NET Core's release index is no longer showing all releases.

splatteredbits:


Whiskey/Functions/Resolve-WhiskeyDotNetSdkVersion.ps1
```diff
@@ -39,10 +39,31 @@ function Resolve-WhiskeyDotNetSdkVersion
...
-        $releasesIndexUri = 'https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json'
+        $urisToTry = @(
+            'https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json',
+            'https://raw.githubusercontent.com/dotnet/core/master/release-notes/releases-index.json'
+        )
+        $releasesIndex = $null
+        foreach( $uri in $urisToTry )
+        {
+            $releasesIndex = 
+                Invoke-RestMethod -Uri $uri -ErrorAction Stop |
```
DecoyJoe: Should the error action here actually be `Ignore`? If Microsoft ends up completely removing the resource, this call will cause an exception and thus the loop will not proceed to trying the other URI's.

Or at least catch a `404` response?

PR #293, opened 2019-10-18
Feature/node module publishing dist tags

### LINKED ISSUE #285: Add support for publishing prerelease versions with the `PublishNodeModule` task
Update the `PublishNodeModule` task to automatically publish prerelease versions of node modules if the build is running on a branch that matches one of the `Prerelease` patterns specified in the `Version` task.

DecoyJoe:
Resolves #285 

* Added a new built-in Whiskey variable `WHISKEY_SEMVER2_PRERELEASE_ID` which contains the prerelease identifier of the prerelease label on a SemVer2 version string. E.g. Given the version `1.2.3-alpha.47+buildmetata`, this variable would be `alpha`.
* When publishing a prerelease version of a node module, the `PublishNodeModule` task will now automatically publish the module with the prerelease identifier as the distribution tag.
* Added a `Tag` property to the `PublishNodeModule` task that controls what distribution tag the module is published with. If specified, this property takes precendence over a tag from a prerelease identifier.

### DIFF
--- a/Test/PublishNodeModule.Tests.ps1
+++ b/Test/PublishNodeModule.Tests.ps1
@@ -7,6 +7,7 @@ $credential = $null
 $context = $null
 $packageJsonPath = $null
 $prerelease = $null
+$testRoot = $null
 $threwException = $false
 
 function Init
@@ -15,14 +16,15 @@ function Init
     $script:context = $null
     $script:packageJsonPath = $null
     $script:prerelease = $null
+    $script:testRoot = New-WhiskeyTestRoot
     $script:threwException = $false
 
-    Install-Node
+    Install-Node -BuildRoot $testRoot
 }
 
 function Reset
 {
-    Remove-Node
+    Remove-Node -BuildRoot $testRoot
 }
 
 function GivenPackageJson
@@ -31,7 +33,7 @@ function GivenPackageJson
         $Content
     )
 
-    $script:packageJsonPath = Join-Path -Path $TestDrive.Fullname -ChildPath 'package.json'
+    $script:packageJsonPath = Join-Path -Path $testRoot -ChildPath 'package.json'
     New-Item -Path $packageJsonPath -ItemType File -Value $Content -Force | Out-Null
 }
 
@@ -221,7 +223,7 @@ function WhenPublishingNodeModule
         $version = '{0}-{1}' -f $version, $prerelease
     }
 
-    $script:context = New-WhiskeyTestContext -ForBuildServer -ForBuildRoot $TestDrive.FullName -ForVersion $version
+    $script:context = New-WhiskeyTestContext -ForBuildServer -ForBuildRoot $testRoot -ForVersion $version
 
     $parameter = @{ }
     if( $WithCredentialID )
--- a/Test/PublishNodeModule.Tests.ps1
+++ b/Test/PublishNodeModule.Tests.ps1
@@ -156,6 +156,41 @@ function ThenPackageJsonVersion
     $packageJson | Select-Object -ExpandProperty 'version' | Should -Be $ExpectedVersion
 }
 
+function ThenPublishedWithTag
+{
+    param(
+        $ExpectedTag
+    )
+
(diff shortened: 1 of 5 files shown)


Test/PublishNodeModule.Tests.ps1
```diff
@@ -156,6 +158,41 @@ function ThenPackageJsonVersion
...
+    )
+
+    Assert-MockCalled   -CommandName 'Invoke-WhiskeyNpmCommand' `
+                        -ModuleName 'Whiskey' `
+                        -Times 1 `
+                        -Exactly `
+                        -ParameterFilter {
+                            if( $Name -ne 'publish')
+                            {
+                                return $false
```
splatteredbits: I've never liked how hard it is to tell what parameter was bad in these filters. Last week, I had the idea to add `Write-Error` statements to these filters so I didn't have to write and enable debug statements. I was going to suggest that, but then I thought "Why not just add `Should` assertions?`

Idid some testing and it looks like you *can* use `Should` assertions inside  parameter filters. I think we should move to a pattern where instead of returning `$false`, we make an assertion. Then, after all the assertions, return `$true`.

```powershell
$Name | Should -Be 'publish' -Because 'Invoke-NpmCommand should be called with publish command.'
'--tag' | Should -BeIn $ArgumentList -Because 'Invoke-NpmCommand should be called with "--tag" argument'
$ExpectedTag | Should -BeIn $ArgumentList -Because 'Invoke-NpmCommand should be called with expected tag argument'
$ArgumentList | Should -HaveCount 2 'Invoke-NpmCommand shouldn''t be called with extra arguments'
return $true
```
```

File: Test/PublishNodeModule.Tests.ps1 [hunk 2]
```diff
@@ -229,8 +272,8 @@ Describe 'PublishNodeModule.when publishing node module' {
         }
 "@
         WhenPublishingNodeModule -WithCredentialID 'NpmCred' `
-                                    -WithEmailAddress 'somebody@example.com' `
-                                    -WithNpmRegistryUri 'http://registry@example.com'
+                                 -WithEmailAddress 'somebody@example.com' `
+                                 -WithNpmRegistryUri 'http://registry@example.com'
         ThenNpmrcCreated -WithEmail 'somebody@example.com' -WithRegistry 'http://registry@example.com'
         ThenNpmPackagesPruned
         ThenNodeModulePublished
```
splatteredbits: Add assertion that we aren't passing any tags to `Invoke-WhiskeyNpmCommand`. Add to whatever tests it makes sense. Or add a `WithNoTag` switch to `ThenModeModulePublished`?

Test/Resolve-WhiskeyVariable.Tests.ps1
```diff
@@ -381,6 +394,15 @@ Describe 'Resolve-WhiskeyVariable.WHISKEY_SEMVER2_PRERELEASE' {
     }
 }
 
+Describe 'Resolve-WhiskeyVariable.WHISKEY_SEMVER2_PRERELEASE_ID' {
```
splatteredbits: Add a test for when the version doesn't have a prerelease identifier.

Test/PublishNodeModule.Tests.ps1
DecoyJoe: This does work well, though only if there is a single call to the mocked command with the expected parameter values. When there are multiple calls to the mocked command with different parameter values, you must mix the two patterns, which I personally don't mind:

```
-ParameterFilter {
    if( $Name -ne 'publish')
    {
        return $false
    }

    '--tag' | Should -BeIn $ArgumentList -Because 'Invoke-NpmCommand should be called with "--tag" argument'
    $ExpectedTag | Should -BeIn $ArgumentList -Because 'Invoke-NpmCommand should be called with expected tag argument'
    $ArgumentList | Should -HaveCount 2 -Because 'Invoke-NpmCommand shouldn''t be called with extra arguments'

    return $true
}
```

DecoyJoe:
Wiki will be updated upon approval and all green ✔️ checks.

PR #294, opened 2019-10-18
I'd like to be able to publish while running builds in parallel.

splatteredbits:
Created a task that does its best to wait for other jobs in a build to complete before continuing to the next build task. There are limitations because AppVeyor's API only returns information for the most recently started builds, but this is better than nothing.

### DIFF
--- a/Test/AppVeyorWaitForBuildJobs.Tests.ps1
+++ b/Test/AppVeyorWaitForBuildJobs.Tests.ps1
@@ -63,10 +63,10 @@ function GivenJob
 
         $job = [pscustomobject]@{ 
             'name' = ('job{0}' -f $jobID)
-            'jobId' = $jobID;
-            'status' = $WithStatus;
-            'checks' = $ThatFinishesAtCheck;
-            'finalStatus' = $WithFinalStatus;
+            'jobId' = $jobID
+            'status' = $WithStatus
+            'checks' = $ThatFinishesAtCheck
+            'finalStatus' = $WithFinalStatus
         } 
 
         if( $ThatHasFinishedProperty )
@@ -116,24 +116,31 @@ function GivenWhiskeyYml
 function Init
 {
     $script:failed = $false
-    $script:testRoot = $TestDrive.FullName
+    $script:testRoot = New-WhiskeyTestRoot
     $script:secret = $null
     $script:secretID = $null
     $script:whiskeyYml = $null
     $script:jobs = $null
     $script:currentJobID = $null
     $script:runByAppVeyor = $null
-    Remove-Item -Path (JOin-Path -Path $testRoot -ChildPath 'whiskey.yml') -ErrorAction Ignore
+    Remove-Item -Path (Join-Path -Path $testRoot -ChildPath 'whiskey.yml') -ErrorAction Ignore
 }
 
 function ThenCheckedStatus
 {
     param(
-        [int]$Times,
-        [switch]$Exactly
+        [Parameter(Mandatory)]
+        [int]$Times
     )
 
-    Assert-MockCalled -CommandName 'Invoke-RestMethod' -ModuleName 'Whiskey' -Times $Times -Exactly:$Exactly
+    # Start-Sleep should always get called one time fewer than Invoke-RestMethod
+    $sleptNumTimes = $Times - 1
+    if( $sleptNumTimes -lt 0 )
+    {
+        $sleptNumTimes = 0
+    }
+    Assert-MockCalled -CommandName 'Start-Sleep' -ModuleName 'Whiskey' -Times $sleptNumTimes -Exactly
+    Assert-MockCalled -CommandName 'Invoke-RestMethod' -ModuleName 'Whiskey' -Times $Times -Exactly
 
     $secret = $script:secret
     Assert-MockCalled -CommandName 'Invoke-RestMethod' `
@@ -142,9 +149,13 @@ function ThenCheckedStatus
                             $Headers['Authorization'] | 
                                 Should -Be ('Bearer {0}' -f $secret) `
                                        -Because 'Invoke-RestMethod should be passed authorization header' 
-                            $PSBoundParameters['Verbose'] | Should -Not -BeNullOrEmpty -Because 'should not show Invoke-RestMethod verbose output'
-                            $PSBoundParameters['Verbose'] | Should -BeFalse -Because 'should not show Invoke-RestMethod verbose output'
-                            $Uri.ToString() | Should -Match '^https://ci\.appveyor\.com/api/projects/.*/.*/builds/\d+$' -Because 'should use api/projects/builds endpoint'
+                            $PSBoundParameters['Verbose'] | 
+                                Should -Not -BeNullOrEmpty -Because 'should not show Invoke-RestMethod verbose output'
+                            $PSBoundParameters['Verbose'] | 
+                                Should -BeFalse -Because 'should not show Invoke-RestMethod verbose output'
+                            $expectedUri = '^https://ci\.appveyor\.com/api/projects/.*/.*/builds/\d+$' 
+                            $Uri.ToString() | 
+                                Should -Match $expectedUri -Because 'should use api/projects/builds endpoint'
                             return $true
                         } `
                       -Times $Times `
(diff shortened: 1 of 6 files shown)


appveyor.yml
```diff
@@ -6,11 +6,6 @@ matrix:
   fast_finish: true
 
 environment:
-  NVM_HOME: C:\Users\appveyor\AppData\Roaming\nvm
-  POWERSHELL_GALLERY_API_KEY:
-    secure: Y7B8n8Ott858qlg1xt1Kz7VnvR59aO963p5dquIEQsMUZpvK/Ip5Rmj7G/jBKM6r
-  GITHUB_ACCESS_TOKEN:
-    secure: cEpuAn4TbJ17Y4WMMsHj46T4DLFMGE4n7xafkbiVQL4jueqc9TGG7NHWg4qbwmt7
```
DecoyJoe: Why were these removed? Are they now being set outside of source control, in the AppVeyor console?

Whiskey/Tasks/AppVeyorWaitForBuildJobs.ps1
```diff
@@ -0,0 +1,99 @@
...
+
+        [TimeSpan]$ReportInterval = '00:01:00',
+
+        [string[]]$InProgressStatus = @('running','queued'),
+
+        [string[]]$SuccessStatus = @('success')
+    )
+
+    Set-StrictMode -Version 'Latest'
+    Use-CallerPreference -Cmdlet $PSCmdlet -SessionState $ExecutionContext.SessionState
```
DecoyJoe: Maybe add an `$ErrorActionPreference = 'Stop'` to this task, so that if anything unexpected happens we fail the build rather than the task short-circuiting continuing with the rest of the publishing tasks.

Test/AppVeyorWaitForBuildJobs.Tests.ps1
```diff
@@ -0,0 +1,474 @@
...
+'@
+        GivenJob -WithStatus 'queued' -ThatFinishesAtCheck 2 -WithFinalStatus 'failed' -ThatHasFinishedProperty
+        GivenJob -Current 
+        GivenLatestBuildIDIsNotThisBuild
+        WhenRunningTask -ErrorAction SilentlyContinue
+        ThenFails
+        $Global:Error | Should -Match 'unable to wait'
+        ThenCheckedStatus -Times 1 -Exactly
+    }
+}
```
DecoyJoe: Add a test when the `ApiKeyID` property is not specified.

Whiskey/Tasks/AppVeyorWaitForBuildJobs.ps1
```diff
@@ -0,0 +1,99 @@
+
+function Wait-WhiskeyAppVeyorBuildJob
+{
+    [CmdletBinding()]
+    [Whiskey.Task('AppVeyorWaitForBuildJobs')]
+    param(
+        [Parameter(Mandatory)]
+        [Whiskey.Context]$TaskContext,
+
+        [Parameter(Mandatory)]
```
DecoyJoe: Setting a task property as mandatory via this attribute currently won't work without additional changes to `Invoke-WhiskeyTask` or `Get-TaskParameter` I think. If someone does not specify this attribute in their `whiskey.yml` then the build will hang as PowerShell prompts the user for this mandatory parameter value.

Whiskey/Tasks/AppVeyorWaitForBuildJobs.ps1
```diff
@@ -0,0 +1,99 @@
...
+
+        $failedJobs = $jobsToCheck | Where-Object { $_.status -notin $SuccessStatus }
+
+        if( $failedJobs )
+        {
+            $suffix = 's'
+            if( ($failedJobs | Measure-Object).Count -gt 1 )
+            {
+                $suffix = 's'
+            }
```
DecoyJoe: Typo, first `$suffix` should be `$suffix = ''`

Test/AppVeyorWaitForBuildJobs.Tests.ps1
```diff
@@ -0,0 +1,474 @@
...
+        {
+            $script:currentJobID = $jobID
+        }
+
+        $job = [pscustomobject]@{ 
+            'name' = ('job{0}' -f $jobID)
+            'jobId' = $jobID;
+            'status' = $WithStatus;
+            'checks' = $ThatFinishesAtCheck;
+            'finalStatus' = $WithFinalStatus;
```
DecoyJoe: Nit: Unnecessary semicolons.

Test/AppVeyorWaitForBuildJobs.Tests.ps1
```diff
@@ -0,0 +1,474 @@
...
+        $Value
+    )
+
+    $script:whiskeyYml = $Value 
+}
+
+function Init
+{
+    $script:failed = $false
+    $script:testRoot = $TestDrive.FullName
```
DecoyJoe: Use `New-WhiskeyTestRoot` ?

Test/AppVeyorWaitForBuildJobs.Tests.ps1
```diff
@@ -0,0 +1,474 @@
...
+    $script:secret = $null
+    $script:secretID = $null
+    $script:whiskeyYml = $null
+    $script:jobs = $null
+    $script:currentJobID = $null
+    $script:runByAppVeyor = $null
+    $script:latestBuildID = $null
+    Remove-Item -Path (JOin-Path -Path $testRoot -ChildPath 'whiskey.yml') -ErrorAction Ignore
```
DecoyJoe: Typo: `JOin-Path`

Test/AppVeyorWaitForBuildJobs.Tests.ps1
```diff
@@ -0,0 +1,474 @@
...
+    ApiKeyID: AppVeyor
+    CheckInterval: 00:00:00.1
+    ReportInterval: 00:00:00.1
+'@
+        GivenJob -WithStatus 'running' -ThatFinishesAtCheck 2 -WithFinalStatus 'success'
+        GivenJob -Current 
+        GivenJob -WithStatus 'running' -ThatFinishesAtCheck 10 -WithFinalStatus 'success'
+        WhenRunningTask
+        ThenSucceeds
+        ThenCheckedStatus -Times 10
```
DecoyJoe: Also add a test that ensures the task is sleeping for the specified interval? It looks like this just ensures the task iterates a number of times while checking jobs.

appveyor.yml
splatteredbits: AppVeyor merges what's in the UI with what's in your `appveyor.yml` so these aren't actually needed here.

Whiskey/Tasks/AppVeyorWaitForBuildJobs.ps1
splatteredbits: 🤔

Test/AppVeyorWaitForBuildJobs.Tests.ps1
splatteredbits: Ack! I like my semicolons in hashtables. From my cold dead hands!

Whiskey/Tasks/AppVeyorWaitForBuildJobs.ps1
splatteredbits: Any task could be used in the Publish portion of a pipeline. Shouldn't we have `$ErrorActionPreference = 'Stop'` in every task for the same reason?

DecoyJoe: Yes, I agree we should have that for all tasks and I actually thought we already did in `Invoke-WhiskeyTask` but turns out we don't. We mostly rely on calling `Stop-WhiskeyTask` to throw an exception during the build. I wasn't sure if there was a historical reason or not why we aren't using `$ErrorActionPreference = 'Stop'` for all tasks.

splatteredbits: Because it's PowerShell and not all errors are errors worth stopping for.

DecoyJoe:
@splatteredbits The Whiskey [Wiki](https://github.com/webmd-health-services/Whiskey/wiki) still needs updating with documentation for the new `AppVeyorWaitForBuildJobs` task.

PR #295, opened 2019-10-22
Add pwsh shebang to Whiskey build.ps1 scripts for *nix systems

DecoyJoe:


Whiskey/build.ps1
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env pwsh
+
```
splatteredbits: These scripts don't have help? :(

How does PowerShell's help work if a script has a shebang? In my experience, any non-doc comment at the top of a script prevents help from working.

DecoyJoe: That's a bug ([PowerShell #5484](https://github.com/PowerShell/PowerShell/issues/5484)), it does support both if you add a blank line between the shebang and the help:
```
#!/usr/bin/env pwsh

<#
.SYNOPSIS
This is the build.ps1 script
#>
```

I'll add help to these `build.ps1` scripts.

PR #311, opened 2019-11-22
GetPowerShellModule

splatteredbits:
Adding support for custom install location, prerelease modules, and importing after installation (fixes issue #277).

### DIFF
--- a/Test/GetPowerShellModule.Tests.ps1
+++ b/Test/GetPowerShellModule.Tests.ps1
@@ -179,7 +179,7 @@ function ThenErrorShouldBeThrown
 
 Describe 'GetPowerShellModule.when given a module Name' {
     AfterEach { Reset }
-    W''
+    It 'should install the lastest version of that module' {
         Init
         Get-Module 'Zip' | Remove-Module -Force
         GivenModule 'Zip'
--- a/Test/GetPowerShellModule.Tests.ps1
+++ b/Test/GetPowerShellModule.Tests.ps1
@@ -139,7 +139,11 @@ function ThenModuleImported
 
     $module = Get-Module -Name $Name 
     $module | Should -Not -BeNullOrEmpty
-    $module.Path | Should -BeLike ('{0}\{1}\{2}\*' -f $testRoot,$TestPSModulesDirectoryName,$Name)
+    # Cross-platform path generation
+    $expectedPath = Join-Path -Path $testRoot -ChildPath $TestPSModulesDirectoryName
+    $expectedPath = Join-Path -Path $expectedPath -ChildPath $Name
+    $expectedPath = Join-Path -Path $expectedPath -ChildPath '*'
+    $module.Path | Should -BeLike $expectedPath
 }
 
 function ThenModuleNotImported
@@ -175,7 +179,7 @@ function ThenErrorShouldBeThrown
 
 Describe 'GetPowerShellModule.when given a module Name' {
     AfterEach { Reset }
-    It 'should install the lastest version of that module' {
+    W''
         Init
         Get-Module 'Zip' | Remove-Module -Force
         GivenModule 'Zip'
--- a/Test/GetPowerShellModule.Tests.ps1
+++ b/Test/GetPowerShellModule.Tests.ps1
@@ -137,7 +137,9 @@ function ThenModuleImported
         [String]$Name
     )
 
-    Get-Module -Name $Name | Should -Not -BeNullOrEmpty
+    $module = Get-Module -Name $Name 
+    $module | Should -Not -BeNullOrEmpty
+    $module.Path | Should -BeLike ('{0}\{1}\{2}\*' -f $testRoot,$PSModulesDirectoryName,$Name)
 }
 
 function ThenModuleNotImported
@@ -281,7 +283,7 @@ Describe 'GetPowerShellModule.when importing module after installation' {
     AfterEach { Reset }
     It 'should import module into global scope' {
         Init
-        Get-Module -Name 'Glob' | Remove-Module -Force
+        Import-Module -Name (Join-Path $PSScriptRoot -ChildPath ('..\{0}\Glob' -f $PSModulesDirectoryName) -Resolve) -Force
         GivenImport
         GivenModule 'Glob'
         WhenTaskRun
--- a/Test/GetPowerShellModule.Tests.ps1
+++ b/Test/GetPowerShellModule.Tests.ps1
@@ -20,15 +20,18 @@ function Init
     $script:context = New-WhiskeyTestContext -ForDeveloper -ForBuildRoot $testRoot
 }
 
+function GivenImport
+{
+    $taskParameter['Import'] = 'true'
+}
+
 function GivenModule
 {
     Param(
         [String]$Module
     )
-    $script:taskParameter['Name'] = $Module
 
-    $script:modulePath = Join-path -Path $context.BuildRoot -ChildPath $PSModulesDirectoryName
-    $script:modulePath = Join-path -Path $modulePath -ChildPath $taskParameter['Name']
+    $taskParameter['Name'] = $Module
 }
 
 function GivenVersion
@@ -36,7 +39,21 @@ function GivenVersion
     param(
         [String]$Version
     )
-    $script:taskParameter['Version'] = $Version
+    $taskParameter['Version'] = $Version
+}
+
+function GivenPath
+{
+    param(
+        $Path
+    )
+
+    $taskParameter['Path'] = $Path
+}
+
+function GivenPrereleaseAllowed
+{
+    $taskParameter['AllowPrerelease'] = 'true'
 }
 
 function GivenNonExistentModule
@@ -54,10 +71,19 @@ function Reset
     Reset-WhiskeyTestPSModule
 }
 
-function WhenPowershellModuleIsRan
+function WhenTaskRun
 {
     [CmdletBinding()]
     param()
+
+
+    $script:modulePath = Join-Path -Path $context.BuildRoot -ChildPath $PSModulesDirectoryName
+    if( $taskParameter['Path'] )
+    {
+        $script:modulePath = Join-Path $context.BuildRoot -ChildPath $taskParameter['Path']
+    }
+    $script:modulePath = Join-Path -Path $modulePath -ChildPath $taskParameter['Name']
+    $script:modulePath = [IO.Path]::GetFullPath($modulePath)
 
     try
     {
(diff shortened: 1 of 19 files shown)


Whiskey/Tasks/GetPowerShellModule.ps1
```diff
@@ -6,39 +6,58 @@ function Get-WhiskeyPowerShellModule
...
-    Write-WhiskeyInfo -Context $TaskContext -Message ('Installing PowerShell module {0} {1}.' -f $TaskParameter['Name'],$module.Version)
-    $moduleRoot = Install-WhiskeyPowerShellModule -Name $TaskParameter['Name'] `
+    $destination = Join-Path -Path $TaskContext.BuildRoot -ChildPath $psModulesDirectoryName
+    if( $Path )
+    {
+        $destination = $Path
+    }
+    $destination = Resolve-Path -Path $destination -Relative -ErrorAction Ignore
+
+    Write-WhiskeyInfo -Context $TaskContext -Message ('Installing PowerShell module {0} {1} to {2}.' -f $Name,$module.Version,$destination)
```
DecoyJoe: Due to the `Resolve-Path` just above, in a clean directory where the module destination directory doesn't exist yet, this message will write `Installing PowerShell module ... to .`

splatteredbits: Currently, `Resolve-WhiskeyPowerShellModule` creates the destination directory, but it might not in the future. I've made the ` to` portion only display if the destination resolves to an actual path.

Whiskey/Functions/Import-WhiskeyPowerShellModule.ps1
```diff
@@ -41,33 +41,28 @@ function Import-WhiskeyPowerShellModule
...
+            Write-WhiskeyDebug -Message ('PSModuleAutoLoadingPreference = "{0}"' -f $PSModuleAutoLoadingPreference)
+            Write-WhiskeyVerbose -Message ('Importing PowerShell module "{0}" from "{1}".' -f $moduleName,$relativePSModulesRoot)
+            $numErrorsBefore = $Global:Error.Count
+            & {
+                $VerbosePreference = 'SilentlyContinue'
+                Import-Module -Name $moduleDir -Global -Force -ErrorAction Stop -Verbose:$false
+            } 4> $null
+            # Some modules (...cough...PowerShellGet...cough...) write silent errors during import. This causes our tests
+            # to fail. I know this is a little extreme.
+            $numErrorsAfter = $Global:Error.Count - $numErrorsBefore
```
pnstickne: Clear global errors prior?

Whiskey/Functions/Install-WhiskeyPowerShellModule.ps1
```diff
@@ -37,27 +37,65 @@ function Install-WhiskeyPowerShellModule
...
+            [Parameter(Mandatory)]
+            [String]$Version
+        )
+
+        if( [Version]::TryParse($Version,[ref]$null) )
+        {
+            return $Version
+        }
+
+        if( $Version -match ('^(\d+\.\d+\.\d+)') )
```
pnstickne: I tend to prefer `[.]` instead of `\.` to reduce slashes.

Whiskey/Whiskey.psd1
```diff
@@ -159,7 +159,10 @@
 
             # ReleaseNotes of this module
             ReleaseNotes = @'
-* Fixed: Whiskey's error view displays too much information sometimes.
+* The `GetPowerShellModule` task now supports installing prerelease versions of modules. Set the `AllowPrerelease` property to `true`.
```
pnstickne: This release note format still gets me because the cumulative change set does not carry forward :|

splatteredbits: We should make a task that makes a cumulative set of release notes.

Whiskey/Functions/Import-WhiskeyPowerShellModule.ps1
splatteredbits: I’ll play around with this approach.

Test/WhiskeyTest.psm1
```diff
@@ -1,4 +1,6 @@
 
+$PSModulesDirectoryName = 'PSModules'
```
DecoyJoe: Should we have this variable name match the one in [Whiskey.psm1](https://github.com/webmd-health-services/Whiskey/blob/develop/Whiskey/Whiskey.psm1#L14), `$powerShellModulesDirectoryName`?

Test/Invoke-WhiskeyBuild.Tests.ps1
```diff
@@ -359,6 +375,35 @@ function WhenRunningBuild
...
 
+function WhenRunningBuildFromBuildPs1
+{
+    [CmdletBinding()]
+    param(
+    )
+
+    $script:threwException = $false
+    $buildPs1Path = Join-Path $testRoot -ChildPath 'build.ps1'
+        @"
```
DecoyJoe: Arbitrary indentation?

Test/Invoke-WhiskeyBuild.Tests.ps1
```diff
@@ -543,3 +588,65 @@ Build:
...
     }
 }
+
+Describe 'Invoke-WhiskeyBuild.when task violates a strict mode rule' {
+    It 'should fail the build' {
+        Init
+        GivenWhiskeyYml @'
+Build:
+- Version:
+    Version: 0.0.0
```
DecoyJoe: Must we always include a `Version` task in whiskey.yml's used for tests? If so, then at least use the more concise format: `- Version: 0.0.0`

Test/Invoke-WhiskeyBuild.Tests.ps1
```diff
@@ -543,3 +588,65 @@ Build:
...
+- CommandNotFoundTask
+'@
+        WhenRunningBuildFromBuildPs1 -ErrorAction SilentlyContinue
+        ThenBuildFailed -WithErrorMessage 'Build\ failed.'
```
DecoyJoe: Escape the period at the end: `Build\ failed\.`

Test/Invoke-WhiskeyBuild.Tests.ps1
```diff
@@ -543,3 +588,65 @@ Build:
...
+
+Describe 'Invoke-WhiskeyBuild.when task fails' {
+    It 'should fail the build' {
+        Init
+        GivenWhiskeyYml @'
+Build:
+- Version:
+    Version: 0.0.0
+- FailingTask:
+    Message: fdsafjkfsdafjdsf
```
DecoyJoe: Maybe make the message an actual readable string so someone can easily tell if the value is correct in `-WithErrorMessage`.

Test/Invoke-WhiskeyBuild.Tests.ps1
```diff
@@ -543,3 +588,65 @@ Build:
...
+
+Describe 'Invoke-WhiskeyBuild.when cmdlet fails because ErrorAction is Stop' {
+    It 'should fail the build' {
+        Init
+        GivenWhiskeyYml @'
+Build:
+- Version:
+    Version: 0.0.0
+- CmdletErrorActionStopTask:
+    Path: ruirwemdsfirewmk
```
DecoyJoe: Maybe make the message an actual readable string so someone can easily tell if the value is correct in `-WithErrorMessage`.

Test/Invoke-WhiskeyBuild.Tests.ps1
```diff
@@ -543,3 +588,65 @@ Build:
...
+        GivenWhiskeyYml @'
+Build:
+- Version:
+    Version: 0.0.0
+- FailingTask:
+    Message: fdsafjkfsdafjdsf
+'@
+        WhenRunningBuildFromBuildPs1 -ErrorAction SilentlyContinue
+        ThenBuildFailed -WithErrorMessage '\bfdsafjkfsdafjdsf\b'
+        $Global:Error | Should -Not -Match 'Build\ failed\.'
```
DecoyJoe: Where does the "Build failed." message come from and why does it not occur in this case?

Test/Invoke-WhiskeyBuild.Tests.ps1
```diff
@@ -543,3 +588,65 @@ Build:
...
+        GivenWhiskeyYml @'
+Build:
+- Version:
+    Version: 0.0.0
+- CmdletErrorActionStopTask:
+    Path: ruirwemdsfirewmk
+'@
+        WhenRunningBuildFromBuildPs1 -ErrorAction SilentlyContinue
+        ThenBuildFailed -WithErrorMessage '\bruirwemdsfirewmk\b'
+        $Global:Error | Should -Not -Match 'Build\ failed\.'
```
DecoyJoe: Where does the "Build failed." message come from and why does it not occur in this case?

Test/GetPowerShellModule.Tests.ps1
```diff
@@ -174,7 +246,46 @@ Describe 'GetPowerShellModule.when called with clean mode' {
...
+        GivenPath 'FubarSnafu'
+        WhenTaskRun
+        ThenModuleInstalled -InDirectory 'FubarSnafu'
+        GivenCleanMode
+        WhenTaskRun
+        ThenModuleNotInstalled -InDirectory 'FubarSnafu'
+    }
+}
+
+Describe 'GetPowerShellModule.when importing module after installation' {
```
DecoyJoe: Add a test for what happens when importing a module that's already imported. Newer version? Older version?

Test/GetPowerShellModule.Tests.ps1
```diff
@@ -174,7 +246,46 @@ Describe 'GetPowerShellModule.when called with clean mode' {
...
+        Init
+        GivenModule 'Whiskey'
+        GivenVersion '0.43.*-*'
+        GivenPrereleaseAllowed
+        WhenTaskRun
+        ThenModuleInstalled -AtVersion '0.43.0-beta1416'
+    }
+}
+
+Describe 'GetPowerShellModule.when installing to custom directory' {
```
DecoyJoe: Add a test that ensures we can install modules at both the default and custom locations during a build?

Whiskey/Functions/Import-WhiskeyPowerShellModule.ps1
splatteredbits: Looks like that works:

```powershell
$Global:Error.Clear()

for( $idx = 0 ; $idx -lt 10 ; $idx++ )
{
    Write-Error $idx -ErrorAction SilentlyContinue
}

$originalErrors = $Global:Error.Clone()

$Global:Error.Clear()

$Global:Error.AddRange($originalErrors)

$Global:Error
```

So, we could move to this pattern:

```powershell
$originalErrors = $Global:Error.Clone()
$Global:Error.Clear()
try
{
    # Do stuff that might write errors we want to clear.
}
finally
{
    $Global:Error.Clear()
    $Global:Error.AddRange($originalErrors)
}
```

Test/WhiskeyTest.psm1
splatteredbits: I wanted them different so that we didn't accidentally use the global variable from the tests in Whiskey (which is what I did accidentally, and tests passed on my machine but failed on the build server; this also uncovered that Whiskey wasn't correctly failing builds in some situations.)

DecoyJoe: Ok I thought that was your reasoning. In that case, maybe make the name more explicit that it's different for tests? `$PSModulesDirectoryNameForTests` ?

Test/Invoke-WhiskeyBuild.Tests.ps1
splatteredbits: I was just trying to avoid the stupid Whiskey version warning message. I think I'll make a new PR to remove it. It's dumb at this point.

Test/Invoke-WhiskeyBuild.Tests.ps1
splatteredbits: This comes from the place in `Invoke-WhiskeyBuild`  that handles terminating errors that don't quite terminate. This is testing that Whiskey handles actually terminating errors.

Test/Invoke-WhiskeyBuild.Tests.ps1
splatteredbits: This comes from the place in `Invoke-WhiskeyBuild`  that handles terminating errors that don't quite terminate. This is testing that Whiskey handles actually terminating errors and doesn't fall through to the backup handler.

Test/GetPowerShellModule.Tests.ps1
splatteredbits: The test after this one handles that case.

splatteredbits:
@DecoyJoe and @pnstickne : Ready for next review.

PR #315, opened 2019-11-26
Pester4 task improvements:

splatteredbits:
* Use named parameters instead of the TaskParameter hashtable.
* Give users the option to *not* run tests in a background job. Some
testing tools don't like that.
* Allow users to pass custom arguments to Pester via an `Arguments`
property.

I tried to get Pester4 to run tests not in a background job by default,
but there was too much weirdness. Plus, it makes sense that tests should
run in an isolated process. In trying to get the Pester4 task to run
tests in scope, I made some changes to test fixtures that are
beneficial, so I'm keeping them.

Created a new `Write-WhiskeyObject` function so tasks can more-easily
write objects to the output in a recognizable format.

### DIFF
--- a/Whiskey/Tasks/Pester4.ps1
+++ b/Whiskey/Tasks/Pester4.ps1
@@ -8,32 +8,30 @@ function Invoke-WhiskeyPester4Task
         [Parameter(Mandatory)]
         [Whiskey.Context]$TaskContext,
 
-        [Parameter(Mandatory)]
-        [hashtable]$TaskParameter
+        [Whiskey.Tasks.ValidatePath(Mandatory)]
+        [String[]]$Path,
+
+        [String[]]$Exclude,
+
+        [int]$DescribeDurationReportCount = 0,
+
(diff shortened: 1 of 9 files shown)


Whiskey/Tasks/Pester4.ps1
```diff
@@ -54,81 +52,109 @@ function Invoke-WhiskeyPester4Task
...
+        $outputFile = Join-Path -Path $outputFileRoot -ChildPath ('pester+{0}.xml' -f [IO.Path]::GetRandomFileName())
+        $Argument['OutputFile'] = $outputFile
+    }
+
+    if( -not $Argument.ContainsKey('OutputFormat') )
+    {
+        $Argument['OutputFormat'] = 'NUnitXml'
+    }
+
+    $Argument | Write-WhiskeyObject -Context $context -Level Verbose -Verbose
```
DecoyJoe: `-Verbose` Left over from debugging?

PR #323, opened 2020-01-16
Fixed: GetPowerShellModule installs package management modules into t…

splatteredbits:
…he custom install directory.


Whiskey/Whiskey.psd1
```diff
@@ -170,6 +170,7 @@
 * Removed all support for old "VersionFrom", "PrereleaseMap", and "Version" properties in the root of your whiskey.yml file. Use Whiskey's `Version` task instead.
 
 ## Improvements and Bug Fixes
+
```
DecoyJoe: Missing release notes

Whiskey/Tasks/GetPowerShellModule.ps1
```diff
@@ -50,7 +50,6 @@ function Get-WhiskeyPowerShellModule
                                               -Version $Version `
                                               -BuildRoot $TaskContext.BuildRoot `
                                               -AllowPrerelease:$AllowPrerelease `
-                                              -Path $fullPath `
```
DecoyJoe: Are we removing the ability to specify a path for where the PowerShell modules will be downloaded to?

Whiskey/Whiskey.psd1
splatteredbits: This is a fix for new functionality that hasn't been released on master yet.

Whiskey/Tasks/GetPowerShellModule.ps1
splatteredbits: No. We still use it when we install. We just don't need it when we find the version to install. Not sure why I ever passed the path to `Resolve-WhiskeyPowerShellModule`.

Whiskey/Whiskey.psd1
DecoyJoe: 👍

Whiskey/Tasks/GetPowerShellModule.ps1
DecoyJoe: 👍

PR #324, opened 2020-01-29
Feature/pester4 pass arguments to test scripts

splatteredbits:


Whiskey/Tasks/Pester4.ps1
```diff
@@ -30,33 +30,85 @@ function Invoke-WhiskeyPester4Task
...
-                            else
-                            {
-                                Write-WhiskeyVerbose -Context $TaskContext -Message ('         {0} -notlike {1}' -f $_,$exclusion)
-                            }
+                            Write-WhiskeyVerbose -Context $TaskContext -Message ('EXCLUDE  {0} -like    {1}' -f $pathItem,$exclusion)
+                            $excluded = $true
+                        }
+                        else
+                        {
+                            Write-WhiskeyVerbose -Context $TaskContext -Message ('         {0} -notlike {1}' -f $pathItem,$exclusion)
```
DecoyJoe: This verbose message is a little confusing. I'm going to be running a build with `-Verbose` and then see a message like:
```
VERBOSE:          something.tests.ps1 -notlike dontrunme.tests.ps1
```
I'll know that statement is true but not what is means or why it's showing me this. What is being conveyed by showing a user this statement (other than it's `true`)? This reads more like a DEBUG message.

splatteredbits: Debug messages are for developers to see to diagnose problems. Verbose messages are for the user to help diagnose problems. If someone's wildcard expressions aren't working, these messages are helpful to a user to figure what is being matched against what so they can get their wildcards correct.

PR #336, opened 2020-08-06
Feature/task out variable

DecoyJoe:
Added a new common task property `OutVariable` for redirecting a task's output to a Whiskey variable.

### DIFF
--- a/Whiskey/Functions/Get-WhiskeyApiKey.ps1
+++ b/Whiskey/Functions/Get-WhiskeyApiKey.ps1
@@ -43,6 +43,15 @@ function Get-WhiskeyApiKey
     }
 
     $secureString = $Context.ApiKeys[$ID]
-    $stringPtr = [Runtime.InteropServices.Marshal]::SecureStringToBSTR($SecureString)
-    return [Runtime.InteropServices.Marshal]::PtrToStringAuto($stringPtr)
+
+    $convertFromSecureString = Get-Command -Name 'ConvertFrom-SecureString'
+    if( $convertFromSecureString.Parameters.ContainsKey('AsPlainText') )
+    {
+        ConvertFrom-SecureString -SecureString $secureString -AsPlainText
+    }
+    else
+    {
+        $stringPtr = [Runtime.InteropServices.Marshal]::SecureStringToBSTR($SecureString)
+        return [Runtime.InteropServices.Marshal]::PtrToStringBSTR($stringPtr)
+    }
 }
--- a/Whiskey/Functions/Invoke-WhiskeyTask.ps1
+++ b/Whiskey/Functions/Invoke-WhiskeyTask.ps1
@@ -169,7 +169,11 @@ function Invoke-WhiskeyTask
 
     [hashtable]$taskProperties = $Parameter.Clone()
     $commonProperties = @{}
-    foreach( $commonPropertyName in @( 'OnlyBy', 'ExceptBy', 'OnlyOnBranch', 'ExceptOnBranch', 'OnlyDuring', 'ExceptDuring', 'WorkingDirectory', 'IfExists', 'UnlessExists', 'OnlyOnPlatform', 'ExceptOnPlatform' ) )
+    foreach( $commonPropertyName in @(
+        'OnlyBy', 'ExceptBy', 'OnlyOnBranch',
+        'ExceptOnBranch', 'OnlyDuring', 'ExceptDuring',
+        'WorkingDirectory', 'IfExists', 'UnlessExists',
+        'OnlyOnPlatform', 'ExceptOnPlatform', 'OutVariable' ) )
     {
         if ($taskProperties.ContainsKey($commonPropertyName))
         {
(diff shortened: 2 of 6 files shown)


Whiskey/Functions/Get-WhiskeyApiKey.ps1
```diff
@@ -43,6 +43,15 @@ function Get-WhiskeyApiKey
     }
 
     $secureString = $Context.ApiKeys[$ID]
-    $stringPtr = [Runtime.InteropServices.Marshal]::SecureStringToBSTR($SecureString)
-    return [Runtime.InteropServices.Marshal]::PtrToStringAuto($stringPtr)
+
+    $convertFromSecureString = Get-Command -Name 'ConvertFrom-SecureString'
+    if( $convertFromSecureString.Parameters.ContainsKey('AsPlainText') )
+    {
+        ConvertFrom-SecureString -SecureString $secureString -AsPlainText
```
splatteredbits, line 50: Doesn't this require `-Force`?

Whiskey/Functions/Invoke-WhiskeyTask.ps1
```diff
@@ -169,7 +169,11 @@ function Invoke-WhiskeyTask
 
     [hashtable]$taskProperties = $Parameter.Clone()
     $commonProperties = @{}
-    foreach( $commonPropertyName in @( 'OnlyBy', 'ExceptBy', 'OnlyOnBranch', 'ExceptOnBranch', 'OnlyDuring', 'ExceptDuring', 'WorkingDirectory', 'IfExists', 'UnlessExists', 'OnlyOnPlatform', 'ExceptOnPlatform' ) )
+    foreach( $commonPropertyName in @(
+        'OnlyBy', 'ExceptBy', 'OnlyOnBranch',
```
splatteredbits, line 173: Group by related parameters?

```
        'OnlyBy',         'ExceptBy',
        'OnlyOnBranch',   'ExceptOnBranch',
        'OnlyDuring',     'ExceptDuring',
        'OnlyOnPlatform', 'ExceptOnPlatform',
        'IfExists', 'UnlessExists',
        'WorkingDirectory', 'OutVariable'
```

splatteredbits: Don't do this now, but maybe instead of adding all these built-in parameters, these become tasks that run a list of tasks if their condition is met?

```
- RunTasks:
    InWorkingDirectory: 'fubar'
    By: BuildServer
    During: Build
    ExceptOnPlatform: macOS
    Tasks:
    - Log:
        Message: I'm in fubar:
```

`Invoke-WhiskeyTask` is already kind of hairy.

Whiskey/Functions/Invoke-WhiskeyTask.ps1
```diff
@@ -259,7 +263,24 @@ function Invoke-WhiskeyTask
                 $parameter.Remove('Debug')
             }
 
-            & $task.CommandName @parameter
+            $outVariable = $commonProperties['OutVariable']
+
+            if ($outVariable)
+            {
+                $taskOutput = & $task.CommandName @parameter
```
splatteredbits, line 270: Maybe `Tee-Object` so people can at least see the output, too?

```
& $task.CommandName @parameter | Tee-Object -Variable 'taskOutput'
```

The only problem with this is `Tee-Object` fails if `$WhatIf` is `True`.

Whiskey/Functions/Get-WhiskeyApiKey.ps1
DecoyJoe: Nope, not `ConvertFrom-SecureString`:
```powershell
> ConvertFrom-SecureString -?

NAME
    ConvertFrom-SecureString

SYNTAX
    ConvertFrom-SecureString [-SecureString] <securestring> [[-SecureKey] <securestring>] [<CommonParameters>]

    ConvertFrom-SecureString [-SecureString] <securestring> [-AsPlainText] [<CommonParameters>]

    ConvertFrom-SecureString [-SecureString] <securestring> [-Key <byte[]>] [<CommonParameters>]
```

Whiskey/Functions/Invoke-WhiskeyTask.ps1
DecoyJoe: I agree `Invoke-WhiskeyTask` is getting unwieldy but that doesn't seem like the right solution. Just looks like another way to call the `Pipeline` task.

Whiskey/Functions/Invoke-WhiskeyTask.ps1
DecoyJoe: > The only problem with this is Tee-Object fails if $WhatIf is True.
Is that a problem here? We don't support running a Whiskey build with `-WhatIf` nor does `Invoke-WhiskeyTask` support `ShouldProcess`.

However, if we did end up using `Tee-Object` then we'd also have to add another common property, `sensitive`. One use I'd have for `OutVariable` is to run a command that gets a secret to pass to another task. I'd need to use `sensitive: true` to suppress that task's output and not leak the secret in the log.

Whiskey/Functions/Invoke-WhiskeyTask.ps1
splatteredbits: Maybe we move this function to `Pipeline` task and rename that task to `RunTasks`?

DecoyJoe: I don't know, I'm still not convinced. What are we really trying to achieve by that? It seems like we'd be making the codebase a bit cleaner at the expense of the `whiskey.yml` syntax.

Whiskey/Functions/Invoke-WhiskeyTask.ps1
DecoyJoe: I tried implementing this with `Tee-Object` and a common `sensitive` property but I didn't like it as it made actually calling the task really hairy with having to support all combinations of those properties. Given that, I opted to forgo that addition as I don't think it's something we need right now.

DecoyJoe:
@splatteredbits Need another review (just a formatting change).

PR #352, opened 2020-12-01
Added a check to not install modules if already globally available.

tatkinsWebMD:


Whiskey/Functions/Install-WhiskeyPowerShellModule.ps1
```diff
@@ -93,57 +93,74 @@ function Install-WhiskeyPowerShellModule
     {
         $modulesRoot = $Path
     }
-    $moduleRoot = Join-Path -Path $modulesRoot -ChildPath $Name
-    $moduleManifestPath = Join-Path -Path $moduleRoot -ChildPath ('{0}\{1}.psd1' -f $moduleVersionDirName,$Name)
 
-    $manifest = $null
-    $manifestOk = $false
-    try
+    if(Test-WhiskeyPowershellModule -Name $Name -Version $Version)
```
DecoyJoe, line 97: Move this check to [Install-WhiskeyTool](https://github.com/webmd-health-services/Whiskey/blob/0b4ef44d0b5cb55a7909d34fb1298aac687dcaa9/Whiskey/Functions/Install-WhiskeyTool.ps1#L128-L140). This function should always install a module to the local build root when it's called. We want to use `Test-WhiskeyPowerShellModule` to determine _if_ we call `Install-WhiskeyPowerShellModule`, not _how_ it works.

Also it means we won't affect the `GetPowerShellModule` task as it calls this function to install modules. That task should always install modules into the local build root, regardless if they're installed globally (at least with its current design). This is actually the reason for the build failure with these changes.

Whiskey/Functions/Test-WhiskeyPowerShellModule.ps1
```diff
@@ -0,0 +1,52 @@
+function Test-WhiskeyPowershellModule
+{
+    <#
+    .SYNOPSIS
+    Checks if a module is installed globally.
+     
+    .DESCRIPTION
+    The `Test-WhiskeyPowerShellModule` function tests if a module is installed globally. If the module is installed globally, it returns true. Otherwise it returns false.
```
DecoyJoe, line 8: Update the docs to describe _how_ the module is searched for globally. E.g. `Get-Module -ListAvailable` is used to search which looks for installed modules in the list of paths found in the [PSModulePath](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_psmodulepath?view=powershell-5.1) environment variable.

Test/Resolve-WhiskeyPowerShellModule.Tests.ps1
```diff
@@ -56,6 +56,16 @@ function GivenModuleDoesNotExist
     Mock -CommandName 'Find-Module' -ModuleName 'Whiskey'
 }
 
+function GivenModuleInstalledGlobally
+{
+    Mock -CommandName 'Test-WhiskeyPowershellModule' -ModuleName 'Whiskey' -MockWith { $true }
+}
+
+function GivenModuleNotInstalledGlobally
```
DecoyJoe, line 64: Rename these functions to `GivenPkgMgmtModulesInstalledGlobally` to be clear about what modules we're expecting to be available globally.

Test/Test-WhiskeyPowerShellModule.Tests.ps1
```diff
@@ -0,0 +1,152 @@
...
+        [Parameter(Mandatory)]
+        [String]$AtVersion
+    )
+
+    $script:globalModules += [pscustomobject]@{
+        'Name' = $Name;
+        'Version' = $AtVersion;
+    }
+
+    $script:globalModules | Should -Not -BeNullOrEmpty
```
DecoyJoe, line 29: Remove this assertion, no need for it here. This function just sets up the test environment.

Test/Test-WhiskeyPowerShellModule.Tests.ps1
```diff
@@ -0,0 +1,152 @@
...
+
+
+function Reset
+{
+    Reset-WhiskeyTestPSModule
+}
+
+function ThenReturnedFalse
+{
+    $output | Should -Be $false
```
DecoyJoe, line 40: Use the `-BeFalse` operator, e.g. `$output | Should -BeFalse`

Test/Test-WhiskeyPowerShellModule.Tests.ps1
```diff
@@ -0,0 +1,152 @@
...
+}
+
+function ThenReturnedTrue
+{
+    $output | Should -Be $true
```
DecoyJoe, line 45: Use the `-BeTrue` operator, e.g. `$output | Should -BeTrue`

Test/Test-WhiskeyPowerShellModule.Tests.ps1
```diff
@@ -0,0 +1,152 @@
...
+
+    if( $ModuleVersion )
+    {
+        $parameter['Version'] = $ModuleVersion
+    }
+
+    if($script:globalModules)
+    {
+        $globalModules = $script:globalModules
+        $globalModules | Should -Not -BeNullOrEmpty
```
DecoyJoe, line 69: Remove this assertion, no need for it here. Assertions go in the `Then` functions.

Test/Test-WhiskeyPowerShellModule.Tests.ps1
```diff
@@ -0,0 +1,152 @@
...
+    $parameter = @{
+        'Name' = $ModuleName;
+    }
+
+    if( $ModuleVersion )
+    {
+        $parameter['Version'] = $ModuleVersion
+    }
+
+    if($script:globalModules)
```
DecoyJoe, line 66: You can remove this if statement and keep the contents. Always Mock `Get-Module` with `$globalModules` which will always either be an empty list or have some mocked modules.

Test/Import-WhiskeyPowerShellModule.Tests.ps1
```diff
@@ -0,0 +1,137 @@
...
+function Init
+{
+    $Global:Error.Clear()
+    $script:testRoot = New-WhiskeyTestRoot
+}
+
+function GivenModuleInstalledLocally
+{
+    Mock -CommandName 'Test-Path' -ModuleName 'Whiskey' -MockWith { $true }
+    Mock -CommandName 'Import-Module' -ModuleName 'Whiskey' -MockWith { $true }
```
DecoyJoe, line 17: Move this `Import-Module` mock to the `Init` function as we should always mock it, we don't need to actually import any modules for the tests.

Test/Import-WhiskeyPowerShellModule.Tests.ps1
```diff
@@ -0,0 +1,137 @@
...
+{
+    Mock -CommandName 'Test-Path' -ModuleName 'Whiskey' -MockWith { $false } 
+}
+
+function GivenModuleNotInstalledGlobally
+{
+    Mock -CommandName 'Test-WhiskeyPowershellModule' -ModuleName 'Whiskey' -MockWith { $false }
+}
+
+function Import-PowerShellModule
```
DecoyJoe, line 36: Rename this function to `WhenImportingPowerShellModule`

Test/Import-WhiskeyPowerShellModule.Tests.ps1
```diff
@@ -0,0 +1,137 @@
...
+}
+
+function Reset
+{
+    Reset-WhiskeyTestPSModule
+}
+
+function ThenImportsLocalModule
+{
+    Assert-MockCalled -CommandName 'Test-Path' -ModuleName 'Whiskey' -Times 1
```
DecoyJoe, line 57: We don't need this assertion, just need to assert that we called `Import-Module` with the expected module name.

Test/Import-WhiskeyPowerShellModule.Tests.ps1
```diff
@@ -0,0 +1,137 @@
...
+    Assert-MockCalled -CommandName 'Import-Module' -ModuleName 'Whiskey' -Times 1
+}
+
+function ThenImportsGlobalModule
+{
+    Assert-MockCalled -CommandName 'Test-WhiskeyPowerShellModule' -ModuleName 'Whiskey' -Times 1
+    Assert-MockCalled -CommandName 'Test-Path' -ModuleName 'Whiskey' -Times 0
```
DecoyJoe, line 64: Remove these assertions, we only need to assert that we called `Import-Module` with the expected module name.

Test/Import-WhiskeyPowerShellModule.Tests.ps1
```diff
@@ -0,0 +1,137 @@
...
+    Assert-MockCalled -CommandName 'Import-Module' -ModuleName 'Whiskey' -Times 1
```
DecoyJoe, line 65: Add a `-ParameterFilter` to this assertion to ensure the expected module got imported, and not just that we called the command.

DecoyJoe (changes requested):

DecoyJoe (changes requested):

Whiskey/Functions/Import-WhiskeyPowerShellModule.ps1
```diff
@@ -8,70 +8,94 @@ function Import-WhiskeyPowerShellModule
     .DESCRIPTION
     The `Import-WhiskeyPowerShellModule` function imports a PowerShell module that is needed/used by a Whiskey task. Since Whiskey tasks all run in the module's scope, the imported modules are imported into the global scope. If a module with the same name is currently loaded, it is removed and re-imported.
 
-    The module must be installed in Whiskey's PowerShell modules directory. Use the `RequiresTool` attribute on a task to have Whiskey install a module in this directory or the `GetPowerShellModule` task to install a module in the appropriate place.
+    If the `InstalledGlobally` switch is set, the module must be installed globally and the path to the module must exist in the PSModulePath environment variable. If multiple versions of the module exist, the latest version will be imported unless a version is provided.
```
DecoyJoe, line 11: There is no `InstalledGlobally` switch on this function.

Test/GetPowerShellModule.Tests.ps1
```diff
@@ -293,12 +293,13 @@ Describe 'GetPowerShellModule.when importing module after installation' {
...
         Import-Module -Name $modulePath -Force
         GivenImport
-        GivenModule 'Glob'
+        GivenModule 'Zip'
+        GivenVersion '0.2.0'
         WhenTaskRun
-        ThenModuleInstalled
-        ThenModuleImported 'Glob'
+        ThenModuleInstalled -AtVersion '0.2.0'
+        ThenModuleImported 'Zip'
```
DecoyJoe, line 303: Also add a test for when `GetPowerShellModule` is given a module name that is installed globally (i.e. in `$PSModulePath`).

Whiskey/Whiskey.psd1
```diff
@@ -163,6 +163,7 @@
             ReleaseNotes = @'
 * Added `InstallNode` task for installing Node.js.
 * Fixed: the ProGetUniversalPackage task ignores and overwrites release and build metadata in the `Version` property.
+* Whiskey now uses global PowerShell modules rather than always installing a private copy in a repository's PSModules folder. Private modules in "PSModules" are always used instead of global modules.
```
DecoyJoe, line 166: Should specifically call out that this changes the behavior of the `GetPowerShellModule` task (in a breaking way?). That task will no longer download the given PowerShell module if it's available globally on the system via the `$PSModulePath` environment variable.

DecoyJoe (changes requested):

Whiskey/Whiskey.psd1
splatteredbits: It still does. When given a `Path`, it will install the module if it is not at that path. See https://github.com/webmd-health-services/Whiskey/pull/352/files?file-filters%5B%5D=.ps1&file-filters%5B%5D=.psd1&file-filters%5B%5D=.yml&hide-deleted-files=true#diff-e86b4c8e4aea7c2ab2e71bfeef1f1d86c4f55e9295b683fae6a089e4c6ff8f93R99-R101 .

Test/GetPowerShellModule.Tests.ps1
splatteredbits: This is handled by `Install-WhiskeyPowerShellModule`. I added this and some other test cases to its test fixture.

DecoyJoe (dismissed):

DecoyJoe (approved):

PR #364, opened 2021-03-19
0.48.0. Install node test failing on master

splatteredbits:


Whiskey/Functions/Invoke-WhiskeyTask.ps1
```diff
@@ -247,7 +244,8 @@ function Invoke-WhiskeyTask
             Invoke-Event -EventName ('Before{0}Task' -f $Name) -Property $taskProperties
 
             Write-WhiskeyVerbose -Context $TaskContext -Message ''
-            $startedAt = Get-Date
+            $TaskContext.StartTask($Name)
```
DecoyJoe, line 247: A little confusing. Seeing `.StartTask($Name)` makes me think this is what's doing the work of running the task, but that's not what it does. Would a better name be `.StartTaskStopwatch($Name)` or `.Stopwatch.StartTask($Name)`?

DecoyJoe (approved):

Whiskey/Functions/Invoke-WhiskeyTask.ps1
splatteredbits: I'll try to think of a better name. But I've been working on this PR for too long now. :(

PR #378, opened 2021-12-20
Various Bug Fixes

splatteredbits:


Assembly/Whiskey/BuildInfo.cs
```diff
@@ -40,6 +44,10 @@ public BuildInfo()
 
         public string ScmCommitID { get; set; }
 
+        public string ScmSourceBranch { get; set; }
+
+        public string ScmSourceCommitID { get; set; }
```
DecoyJoe, line 49: I've been wondering if we should make the Whiskey context object be a [PowerShell Class](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_classes) instead of a C# assembly so we can avoid the PowerShell assembly file locking issue on Windows.

Whiskey/Whiskey.psd1
```diff
@@ -155,7 +155,7 @@
             # IconUri = ''
 
             # Any prerelease to use when publishing to a repository.
-            Prerelease = ''
+            Prerelease = 'testing53'
```
DecoyJoe, line 158: Did you mean to check this in?

DecoyJoe (approved):

Whiskey/Whiskey.psd1
splatteredbits: No.

Assembly/Whiskey/BuildInfo.cs
splatteredbits: I played around with classes once. If you want other modules to use them, they can't without a lot of effort or embedding your module in theirs, if I remember correctly. I'd have to play around with it again.

PR #399, opened 2022-08-11
Bugfix: Pester task does not fail when run in background job

splatteredbits:


Whiskey/Tasks/Pester.ps1
```diff
@@ -147,18 +146,40 @@ function Invoke-WhiskeyPesterTask
...
-        Invoke-Pester -Configuration $pesterConfiguration
+        try
+        {
+            $LASTEXITCODE = 0
+            Invoke-Pester -Configuration $pesterConfiguration
+        }
+        finally
+        {
+            Write-Debug "Pester  LASTEXITCODE  $($LASTEXITCODE)"
+            $LASTEXITCODE | Set-Content -Path $ExitCodePath
```
DecoyJoe, line 157: I wonder, is it better to use `$LASTEXITCODE` or `$?`

DecoyJoe (approved):

Whiskey/Tasks/Pester.ps1
splatteredbits: In my testsing, `$?` is `$false` only if Pester is explicitly told to exit with an error or throw an error. If someone forgets to add that configuration to their whiskey.yml file, their builds won't fail if any tests fail. `Invoke-Pester` always exits with a non-zero exit code if any tests fail, so its a little more reliable.

DecoyJoe: 👍

PR #403, opened 2022-11-04
Add .NET SDK roll forward Support

CameronSWilliamson:


Whiskey/Functions/Resolve-WhiskeyDotNetSdkVersion.ps1
```diff
@@ -77,39 +259,236 @@ function Resolve-WhiskeyDotNetSdkVersion
...
-            $sdkVersions |
-            ForEach-Object { $_ -as [Version] } |
-            Where-Object { $_ -like $Version } |
-            Sort-Object -Descending |
-            Select-Object -First 1
+        switch ($RollForward)
+        {
+            Patch 
+            {
+                $resolvedVersion =
```
decoyjoe, line 302: We're using a lot of logic and text parsing here with these helper functions to create custom version objects when we could vastly simplify this with using the built in `[System.Version]` type. The .NET SDK version is a 4-part version number (formatted as a 3-part version) and the `[System.Version]` type is also a 4-part version number.

Parse all the available .NET SDK versions as a 4-part version, cast it as a `[Version]` type, and then sort them (.NET knows how to properly sort `[Version]` objects):

```powershell
$sortedSdkVersions =
    $sdkVersions |
    ForEach-Object {
        $_ -match '^(\d+)\.(\d+)\.(\d{1})(\d+)' | Out-Null
        [Version] "$($Matches[1]).$($Matches[2]).$($Matches[3]).$($Matches[4])"
    } |
    Sort-Object -Descending
```

Now in each block of your switch statement you just call `Where-Object` on that list of versions with varying amounts of conditions and then select the first one.


```powershell
switch ($RollForward)
{
    Patch 
    {
        $resolvedVersion = 
            $sortedSdkVersions |
            Where-Object {
                $_.Major -eq $desiredMajor -and
                $_.Minor -eq $desiredMinor -and
                $_.Build -eq $desiredFeature
            } |
            Select-Object -First 1
    }
}
```

Then you can convert that resolved version back into a 3-part version that the dotnet install script expects, remembering to format the patch version with 2 digits.

```powershell
'{0}.{1}.{2}{3:00}' -f $resolvedVersion.Major, $resolvedVersion.Minor, $resolvedVersion.Build, $resolvedVersion.Revision
```

Also, for the rollForward strategies of `patch`, `feature`, `minor`, and `major` [the docs](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json#rollforward) say that these strategies will attempt to use that version if it's installed and will only roll forward as a fallback. So we need to also determine what versions of the .NET SDK are installed globally and use one of those if it matches the given version and if not only then fallback to searching for a newer one.

[Take this code](https://github.com/webmd-health-services/Whiskey/blob/master/Whiskey/Functions/Install-WhiskeyDotNetSdk.ps1#L38-L54) and put it in a new `Get-InstalledDotNetSdk` function that we'll call here to resolve the currently installed versions.

decoyjoe (changes requested):

Whiskey/Functions/Resolve-WhiskeyDotNetSdkVersion.ps1
```diff
@@ -100,22 +73,22 @@ function Resolve-WhiskeyDotNetSdkVersion
                 Select-Object -ExpandProperty 'version'
         }
 
-        $resolvedVersion =
+        $sortedSdkVersions =
             $sdkVersions |
-            ForEach-Object { $_ -as [Version] } |
-            Where-Object { $_ -like $Version } |
-            Sort-Object -Descending |
-            Select-Object -First 1
```
decoyjoe, line 108: We still need this because `$Version` accepts wildcards. Before we start checking versions against `$RollForward`, do this:

```powershell
if ([WildcardPattern]::ContainsWildcardCharacters($Version))
{
    return $sdkVersions |
           Where-Object { $_ -like $Version } |
           Sort-Object -Descending |
           Select-Object -First 1
}
```

Make sure `$sdkVersions` is already an array of `[Version]` objects.

Whiskey/Functions/Resolve-WhiskeyDotNetSdkVersion.ps1
```diff
@@ -133,7 +106,277 @@ function Resolve-WhiskeyDotNetSdkVersion
...
+
+        $releasesJson.releases |
+            Where-Object { $_ | Get-Member -Name 'sdks' } |
+            Select-Object -ExpandProperty 'sdks' |
+            Select-Object -ExpandProperty 'version'
+    }
+
+    $desiredVersion = $null
+    $sortedSdkVersions = $null
+    if ( $Version -match '^(\d+)\.(\d+)\.(\d{1})(\d+)' )
```
decoyjoe, line 173: Invert this condition to `if ( $Version -notmatch '^(\d+)\.(\d+)\.(\d{1})(\d+)' )` and write an error if it's true:

```powershell
$msg = ".NET SDK version "$($Version)" is invalid. The SDK version must be in the form of a 3-part version number. See https://learn.microsoft.com/en-us/dotnet/core/versions/ for more information."
Write-WhiskeyError -Message $msg
return
```

Whiskey/Functions/Resolve-WhiskeyDotNetSdkVersion.ps1
```diff
@@ -133,7 +106,277 @@ function Resolve-WhiskeyDotNetSdkVersion
...
+    {
+        Disable
+        {
+            $resolvedVersion =
+                $sdkVersions |
+                Where-Object { $_ -like $Version } |
+                Sort-Object -Descending |
+                Select-Object -First 1
+        }
+        Patch
```
decoyjoe, line 210: Make `Patch`, `Feature`, `Minor`, and `Major` all behave the same way: They find the latest patch level for the specified major, minor, and feature band.

Whiskey/Functions/Resolve-WhiskeyDotNetSdkVersion.ps1
```diff
@@ -133,7 +106,277 @@ function Resolve-WhiskeyDotNetSdkVersion
...
+            if ( -not $resolvedVersion )
+            {
+                $resolvedVersion = $sortedSdkVersions | Select-Object -First 1
+            }
+        }
+        LatestPatch
+        {
+            $sortedSdkVersions =
+                $sortedSdkVersions |
+                Where-Object { $_.Build -eq $desiredVersion.Build }
```
decoyjoe, line 305: This should filter versions by the first 3 parts of the version:

```powershell
Where-Object {
  $_.Major -eq $desiredMajor -and
  $_.Minor -eq $desiredMinor -and
  $_.Build -eq $desiredBuild
}
```

Whiskey/Functions/Resolve-WhiskeyDotNetSdkVersion.ps1
```diff
@@ -133,7 +106,277 @@ function Resolve-WhiskeyDotNetSdkVersion
...
+                $resolvedVersion = $sortedSdkVersions | Select-Object -First 1
+            }
+        }
+        LatestFeature
+        {
+            $sortedSdkVersions =
+                $sortedSdkVersions
+            $resolvedVersion =
+                $sortedSdkVersions |
+                Where-Object { $_ -in $installedVersions }
```
decoyjoe, line 321: This should filter by the first 2 parts of the version:

```powershell
Where-Object {
  $_.Major -eq $desiredMajor -and
  $_.Minor -eq $desiredMinor
}
```

decoyjoe (changes requested):

Whiskey/Functions/Resolve-WhiskeyDotNetSdkVersion.ps1
CameronSWilliamson: I decided to omit the first 2 parts of the version because when we grab the .NET SDK versions from Microsoft we are only able to target a specific major-minor set. These can be still be added for clarity but they are unnecessary for both the `Patch` and `Feature` branches

decoyjoe: Ah that makes sense now. I would still add them for clarity as it can be unclear, evidenced by comment.

Whiskey/Functions/Resolve-WhiskeyDotNetSdkVersion.ps1
CameronSWilliamson: The `Minor` and `Major` versions need to operate in a slightly different way than the `Feature` and `Patch` versions. As mentioned below, we need to grab the available SDK versions each time we try to increment either a `Major` or a `Minor` version.

decoyjoe: That is the fallback behavior when a SDK version is not installed. Here we are resolving a version to download so we aren't worried about needing to fallback. All these strategies initially behave in the same way, other than how they handle fallback. This is what they do per the [global.json docs](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json#rollforward) (omitting the fallback verbage):

> Uses the latest patch level for the specified major, minor, and feature band.

That's exactly what we're going to do here. Find the latest patch level for the specified `major`, `minor`, and `feature` versions when the strategy is `major`, `minor`, `feature`, or `patch`.

Test/Install-WhiskeyDotNetTool.Tests.ps1
```diff
@@ -305,16 +387,25 @@ try
...
+        }
+    }
+
+    Describe 'Install-WhiskeyDotNetTool.when installing DotNet and patch lower is already installed' {
+        It 'should use global.json''s version' {
+            Init
+            GivenGlobalDotNetInstalled '2.1.514'
+            GivenGlobalJsonRollForwardAndSdkVersion -Version '2.1.500' -RollForward Patch
+            GivenGlobalDotNetHasValidVersion -Version '2.1.514'
+            WhenInstallingDotNetTool
```
decoyjoe, line 404: This test has no assertions. It should be asserting that:
* The function returns `'dotnet'`.
* That `Install-WhiskeyDotNetSdk` is never called.

Test/Install-WhiskeyDotNetTool.Tests.ps1
```diff
@@ -252,24 +323,36 @@ Describe 'Install-WhiskeyDotNetTool.when existing global.json contains invalid J
...
         WhenInstallingDotNetTool
         ThenReturnedValidDotNetPath
-        ThenGlobalJsonVersion (Get-DotNetLatestLtsVersion)
-        ThenDotNetSdkVersion (Get-DotNetLatestLtsVersion)
+        ThenVersionToInstall -ExpectedVersion (Get-DotNetLatestLtsVersion)
+    }
+}
+
+
+Describe 'Install-WhiskeyDotNetTool.when installing DotNet and global.json has Patch rollforward in global.json' {
```
decoyjoe, line 345: Add a test for when `rollForward` is an invalid value.

Test/Resolve-WhiskeyDotnetSdkVersion.Tests.ps1
```diff
@@ -210,3 +245,176 @@ Describe 'Resolve-WhiskeyDotNetSdkVersion.when Microsoft moves the index again'
         ThenReturnedNothing
     }
 }
+
+Describe 'Resolve-WhiskeyDotNetSdkVersion.when RollForward is diabled and patch is found' {
```
decoyjoe, line 249: Typo: `diabled`

Test/Resolve-WhiskeyDotnetSdkVersion.Tests.ps1
```diff
@@ -210,3 +245,176 @@ Describe 'Resolve-WhiskeyDotNetSdkVersion.when Microsoft moves the index again'
...
+        WhenResolvingSdkVersion -Version '3.0.000' -RollForward Disable
+        ThenError -Message 'not be found'
+    }
+}
+
+Describe 'Resolve-WhiskeyDotNetSdkVersion.when RollForward is patch and patch is found' {
+    It 'should use specified version' {
+        Init
+        GivenReleases '3.0'
+        GivenSDKVersions '3.0.000', '3.0.100' -ForRelease '3.0'
```
decoyjoe, line 273: You need multiple eligible versions here to ensure it's choosing the correct one, which should be the latest patch.

```powershell
GivenSDKVersions '3.0.000', '3.0.001', '3.0.100' -ForRelease '3.0'
WhenResolvingSdkVersion -Version '3.0.000' -RollForward Patch
ThenResolvedVersion '3.0.001'
```

Test/Resolve-WhiskeyDotnetSdkVersion.Tests.ps1
```diff
@@ -210,3 +245,176 @@ Describe 'Resolve-WhiskeyDotNetSdkVersion.when Microsoft moves the index again'
...
+    It 'should fail' {
+        Init
+        GivenReleases '3.0'
+        GivenSDKVersions '3.0.100' -ForRelease '3.0'
+        WhenResolvingSdkVersion -Version '3.0.000' -RollForward Disable
+        ThenError -Message 'not be found'
+    }
+}
+
+Describe 'Resolve-WhiskeyDotNetSdkVersion.when RollForward is patch and patch is found' {
```
decoyjoe, line 269: Create a loop to iterate over the same set of tests for the `patch`, `feature`, `minor`, `major` strategies since they are supposed to all behave the same.

```powershell
foreach ($strategy in @('patch', 'feature', 'minor', 'major'))
{
    Describe "Resolve-WhiskeyDotNetSdkVersion.when RollForward is $($strategy) and exact version does exist" {
        It 'should resolve latest patch' {
            Init
            GivenReleases '3.0'
            GivenSDKVersions '3.0.000', '3.0.001', '3.0.100' -ForRelease '3.0'
            WhenResolvingSdkVersion -Version '3.0.000' -RollForward $strategy
            ThenResolvedVersion '3.0.001'
        }
    }

    Describe "Resolve-WhiskeyDotNetSdkVersion.when RollForward is $($strategy) and exact version does not exist" {
        It 'should resolve latest patch' {
            Init
            GivenReleases '3.0'
            GivenSDKVersions '3.0.001', '3.0.002' -ForRelease '3.0'
            WhenResolvingSdkVersion '3.0.000' -RollForward $strategy
            ThenResolvedVersion '3.0.002'
        }
    }
}
```

decoyjoe (changes requested):

Whiskey/Functions/Resolve-WhiskeyDotNetSdkVersion.ps1
```diff
@@ -22,118 +24,231 @@ function Resolve-WhiskeyDotNetSdkVersion
...
+    $releasesIndex =
+        $releasesIndex |
+        Where-Object { [Version]::TryParse($_.'channel-version', [ref]$null) } |
+        ForEach-Object {
+            $_.'channel-version' = [Version]$_.'channel-version'
+            $_
+        } |
+        Sort-Object -Property 'channel-version' -Descending
+
+    # $Version -match '^\d+\.(?:\d+|\*)|^\*' | Out-Null
```
decoyjoe, line 105: If this is no longer needed then delete the line.

Whiskey/Functions/Resolve-WhiskeyDotNetSdkVersion.ps1
```diff
@@ -22,118 +24,231 @@ function Resolve-WhiskeyDotNetSdkVersion
...
 
-        $sdkVersions = & {
-            $releasesJson.releases |
-                Where-Object { $_ | Get-Member -Name 'sdk' } |
-                Select-Object -ExpandProperty 'sdk' |
-                Select-Object -ExpandProperty 'version'
+    $release = $releasesIndex |
+        Sort-Object -Property 'channel-version' -Descending |
+        Where-Object { $_.'channel-version' -like $matcher } |
+        Select-Object -First 1
```
decoyjoe, line 132: Nit: Multi-line formatting

```powershell
$release =
    $releasesIndex |
    Sort-Object -Property 'channel-version' -Descending |
    Where-Object { $_.'channel-version' -like $matcher } |
    Select-Object -First 1
```

Test/Install-WhiskeyDotNetTool.Tests.ps1
```diff
@@ -19,6 +19,8 @@ function Init
     $script:globalDotNetDirectory = Join-Path -Path $TestDrive.FullName -ChildPath 'GlobalDotNetSDK'
     $script:version = $null
     $script:workingDirectory = $null
+    Mock -CommandName 'Get-Command' -ModuleName 'Whiskey' -ParameterFilter { $Name -eq 'dotnet' } -MockWith { $null }.GetNewClosure()
```
decoyjoe, line 22: You can actually just leave off the `-Mockwith` parameter and the mock will by default return `$null`.

Test/Install-WhiskeyDotNetTool.Tests.ps1
```diff
@@ -19,6 +19,8 @@ function Init
     $script:globalDotNetDirectory = Join-Path -Path $TestDrive.FullName -ChildPath 'GlobalDotNetSDK'
     $script:version = $null
     $script:workingDirectory = $null
+    Mock -CommandName 'Get-Command' -ModuleName 'Whiskey' -ParameterFilter { $Name -eq 'dotnet' } -MockWith { $null }.GetNewClosure()
+    Mock -CommandName 'dotnet' -ModuleName 'Whiskey' -MockWith { 'return some strings'; $global:LASTEXITCODE = 1 }.GetNewClosure()
```
decoyjoe, line 23: A cleaner way to mock the exit code is to run a process that actually exits with the desired exit code:

```powershell
Mock -CommandName 'dotnet' -ModuleName 'Whiskey' -MockWith { cmd /c exit 1 }
```

Test/Install-WhiskeyDotNetTool.Tests.ps1
```diff
@@ -252,69 +286,100 @@ Describe 'Install-WhiskeyDotNetTool.when existing global.json contains invalid J
...
-            ThenDotNetLocallyInstalled '2.1.505'
-        }
+
+Describe 'Install-WhiskeyDotNetTool.when installing DotNet and global.json has Patch rollforward in global.json' {
+    It 'should install latest patch for version' {
+        Init
+        GivenGlobalJson -Version '2.1.500' -RollForward Patch
+        GivenDotNetSuccessfullyInstalls
+        WhenInstallingDotNetTool
+        # ThenVersionToInstall -ExpectedVersion '2.1.526'
```
decoyjoe, line 314: This assertion is commented out.

Test/Install-WhiskeyDotNetTool.Tests.ps1
```diff
@@ -252,69 +286,100 @@ Describe 'Install-WhiskeyDotNetTool.when existing global.json contains invalid J
...
+    It 'should use global version' {
+        Init
+        GivenGlobalDotNetInstalled '2.1.505'
+        GivenGlobalDotNetHasValidVersion -Version '2.1.505'
+        GivenVersion '2.1.505'
+        GivenDotNetSuccessfullyInstalls
+        WhenInstallingDotNetTool
+        ThenDotNetLocallyInstalled '2.1.505'
+        ThenReturnedValidDotNetPath
+        ThenDotNetWasInstalled
```
decoyjoe, line 355: If the scenario is `'Install-WhiskeyDotNetTool.when specified version of DotNet exists globally'` then this should be testing the opposite, that Whiskey _**did not**_ install a local version.

decoyjoe (changes requested):

decoyjoe (approved):

PR #423, opened 2024-03-29
Bugfix/auto increment prerelease versions

decoyjoe:


Test/Version.Tests.ps1
```diff
@@ -446,6 +446,27 @@ Describe 'Version' {
...
+                        summary = 'Not Found';
+                    }
+                } | ConvertTo-Json
+             }
+
+        GivenFile 'package.json' '{ "name": "this-package-has-not-been-published-yet", "version": "0.0.0-rc.0" }'
+        GivenProperty @{ Path = 'package.json'; IncrementPrereleaseVersion = $true; }
+        WhenRunningTask
+        ThenVersionIs '0.0.0'
+        ThenSemVer1Is '0.0.0-rc1'
```
splatteredbits, line 466: It should use the original prerelease version number, so should be `0.0.0-rc.0` and `0.0.0-rc0`, since that's what's in the package.json file.

splatteredbits (changes requested):

Test/Version.Tests.ps1
decoyjoe: You're right. Fixed.

decoyjoe:
@splatteredbits This is ready for another review.

splatteredbits (approved):

PR #425, opened 2024-06-25
Fix: Disallow publishing of node modules with the same version.

KhoiKy:


CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 
 # Whiskey Changelog
 
+## 0.61.1
+
+> Released 25 Jun 2024
+
+### Fixed
+
+* The `PublishNodeModule` task will no longer allow publishing of node modules while retaining the same version.
```
splatteredbits, line 12: Reword to "node modules with the same version".

splatteredbits (approved):

Test/PublishNodeModule.Tests.ps1
```diff
@@ -297,13 +303,41 @@ Describe 'PublishNodeModule' {
...
+        {
+            "name": "publishnodemodule_test",
+            "version": "1.2.0"
+        }
+"@
+        WhenPublishingNodeModule -WithCredentialID 'NpmCred' `
+                                 -WithEmailAddress 'somebody@example.com' `
+                                 -WithNpmRegistryUri 'http://registry@example.com' `
+                                 -ErrorAction SilentlyContinue
+        ThenTaskFailed 'failed with exit code 1'
```
decoyjoe, line 325: I don't understand this test. What version is expected/where does it come from? Why does this scenario fail the task?

Whiskey/Tasks/PublishNodeModule.ps1
```diff
@@ -89,19 +89,19 @@ function Publish-WhiskeyNodeModule
 
         Copy-Item -Path $originalPackageJsonPath -Destination $backupPackageJsonPath
         Invoke-WhiskeyNpmCommand -Name 'version' `
-                                 -ArgumentList $TaskContext.Version.SemVer2NoBuildMetadata, '--no-git-tag-version', '--allow-same-version' `
+                                 -ArgumentList $TaskContext.Version.SemVer2NoBuildMetadata, '--no-git-tag-version' `
```
decoyjoe, line 92: I don't understand how this fixes the bug. Isn't the issue with the `npm publish` command?

decoyjoe (changes requested):

Whiskey/Tasks/PublishNodeModule.ps1
KhoiKy: No the issue is with the `npm version` command which is responsible for bumping a package version: https://docs.npmjs.com/cli/v10/commands/npm-version#allow-same-version

Test/PublishNodeModule.Tests.ps1
KhoiKy: Look at the update I made to the `WhenPublishingNodeModule` function for the tests.

Previously the way the tests were set up, the version is retrieved from the package.json that's created from the `GivenPackageJson` function. I've added a new parameter `WithVersion` so we can explicitly set the version for the Whiskey Task Context.

I'll update the tests so that they all specify a version explicitly on the `WithVersion` parameter so that it's easier to understand.

decoyjoe: The test seems flawed, it's not representative of how the version would be resolved during a real build.

A `Version` task would read the version from the `package.json` file, which would mean the version in the TaskContext object matches the package.json version. But in the test, you're manually changing the TaskContext version to something else.

That's exactly why the test behaves in this way:

> Previously the way the tests were set up, the version is retrieved from the package.json that's created from the GivenPackageJson function.

because that's how the build would get its version.

Whiskey/Tasks/PublishNodeModule.ps1
decoyjoe: The `npm version` command just updates the version number in the package.json file on-disk. I don't think it knows anything about what versions have already been published to the registry.

KhoiKy: Ok I understand now. I'll use the `npm view` command to get latest published version of a package and compare that to the version in the Task Context. That should do the job.

KhoiKy: Turns out this was an issue with the setting in our private node module registry that allowed over writing of packages. Closing out PR.

PR #428, opened 2024-09-24
Deprecate PublishBitbucketServerTag task

splatteredbits:
Should we actually do this?


decoyjoe:
Tagging should be done by whatever tool is doing the publishing, since that's the tool that's aware that publishing occurred. If Whiskey is going to continue to publish artifacts then I think it makes sense to keep this task around.

splatteredbits:
Turns out we do use and need this task.

PR #457, opened 2026-03-30
Treat PR builds as if they're running on the source branch

splatteredbits:


Test/New-WhiskeyContext.Tests.ps1
```diff
@@ -80,7 +80,7 @@ BeforeAll {
 
             $BuildNumber = '1',
 
-            [String] $IsPRFromBranch
+            [switch] $IsPR
```
KhoiKy: nit: This probably isn't used anymore and can be removed.

KhoiKy (approved):

PR #459, opened 2026-04-20
Fixed: Whiskey fails to import an already installed PowerShell module…

splatteredbits:
… if it is a prerelease version. Also fixed: tasks can't request prerelease versions of modules to install/use.


Whiskey/Functions/Install-WhiskeyPowerShellModule.ps1
```diff
@@ -34,7 +35,9 @@ function Install-WhiskeyPowerShellModule
         [String]$Version,
 
         [Parameter(Mandatory)]
-        # Modules are saved into a PSModules directory. This is the directory where PSModules directory should created, *not* the path to the PSModules directory itself, i.e. this is the path to the "PSModules" directory's parent directory.
+        # Modules are saved into a PSModules directory. This is the directory where PSModules directory should created,
```
decoyjoe, line 38: Thought for the future: Whiskey should probably save the modules that it uses internally in a separate `.whiskey/PSModules` directory to avoid polluting the root `PSModules` directory which is used for user requested modules and by Prism.

decoyjoe (approved):

Whiskey/Functions/Install-WhiskeyPowerShellModule.ps1
splatteredbits: Agreed. I had the same thought, but decided not to do that here. I was going to put them in `.output/PSModules`, but `.whiskey/PSModules` is probably a better place.

PR #460, opened 2026-04-29
Fix: Build metadata uses the pull request's transient merge commit ID instead of the source branch's commit ID

KhoiKy:


Whiskey/Functions/Get-WhiskeyBuildMetadata.ps1
```diff
@@ -60,9 +60,18 @@ function Get-WhiskeyBuildMetadata
         $buildInfo.JobUri = Get-EnvironmentVariable 'JOB_URL'
         $buildInfo.ScmUri = Get-EnvironmentVariable 'GIT_URL'
         $buildInfo.ScmCommitID = Get-EnvironmentVariable 'GIT_COMMIT'
+        $scmBranch = Get-EnvironmentVariable 'GIT_BRANCH'
+
         # For PR builds, GIT_BRANCH gets set to `PR-ID` where ID is a the PR number and CHANGE_BRANCH is the source
         # branch, so always use CHANGE_BRANCH if it exists.
-        $scmBranch = Get-EnvironmentVariable 'CHANGE_BRANCH','GIT_BRANCH'
+        if (Test-Path -Path 'env:CHANGE_BRANCH')
```
splatteredbits, line 66: There's got to be an env variable to use instead.

splatteredbits: Change this to:

```
        $scmBranch = Get-EnvironmentVariable 'CHANGE_BRANCH','GIT_BRANCH'
        if (Test-Path -Path 'env:CHANGE_BRANCH')
        {
            # In a PR build, Jenkins doesn't provide the commit ID of the source branch.
            $buildInfo.ScmCommitID = git rev-parse $scmBranch
        }
```

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 
 # Whiskey Changelog
 
+## 0.65.3
+
+Fixed: Build metadata uses the pull request's transient merge commit ID instead of the source branch's commit ID.
```
splatteredbits: Rephrase to `Fixed: In Jenkins, the commit ID for PR builds is set to the transient merge commit ID instead of the source branch's commit ID.`

splatteredbits (changes requested):

Whiskey/Functions/Get-WhiskeyBuildMetadata.ps1
```diff
@@ -60,9 +60,18 @@ function Get-WhiskeyBuildMetadata
...
+
         # For PR builds, GIT_BRANCH gets set to `PR-ID` where ID is a the PR number and CHANGE_BRANCH is the source
         # branch, so always use CHANGE_BRANCH if it exists.
-        $scmBranch = Get-EnvironmentVariable 'CHANGE_BRANCH','GIT_BRANCH'
+        if (Test-Path -Path 'env:CHANGE_BRANCH')
+        {
+            $scmBranch = Get-EnvironmentVariable 'CHANGE_BRANCH'
+
+            # Populate Commit ID from source branch
+            $buildInfo.ScmCommitID = git rev-parse $scmBranch
```
decoyjoe: I need to learn more about what we're fixing here. If there is a merge commit we do indeed want the merge commit, not the source branch commit.

What's the scenario that we're bug fixing?

splatteredbits: If the build is a PR, Jenkins' `COMMIT_ID` environment variable is the commit ID of the transient merge commit Jenkins creates before it runs a build. This is something I missed when I removed Whiskey's PR handling. During a PR, Whiskey needs to act like the build is happening on the source branch.

decoyjoe: In that case, I don't think we should be tagging any commit during a PR build, because the commit being built is a temporary merge commit that won't exist after the build finishes. There is no commit to tag which points to that exact build.

The PR build merges the target branch into the source branch before building, so it can produce different output than a build of the source branch alone would. For example, if `main` has changes that affect the source branch's behavior, the PR build reflects the merged result, while a build of the source branch commit itself would not. Tagging the source commit would therefore associate the tag with artifacts that weren't actually built from it.

splatteredbits: 🤔

splatteredbits: Now I'm not sure how Whiskey should handle pull requests at all. :(

AppVeyor and Jenkins behave differently with respect to pull requests. Jenkins creates a custom branch and a transient commit. AppVeyor uses the target branch as the branch name and a transient commit on that branch.

I updated Whiskey so it uses the PR's source branch as the branch name of the current build. If I had known about the transient commit ID, I would have also updated Whiskey to use the source branch's commit as the current build's commit ID. I think this is what people expect. It's what I expected. Even though the build server is creating a transient commit, the build was still triggered and run from a specific commit in a repository. As a developer, I want to know what commit corresponds to a build. I see the transient commit as an implementation detail of the build server. I think that's what Whiskey should reflect.

Do we revert this change and treat the branch and commit of a PR build differently?

Do we remove PR handling completely from Whiskey and just use the branch and commit as given by the build server? Internally, we'd have to add `PR-*` to the list of branches to publish on in each repo's whiskey.yml file and a `.UnlessExists: env:CHANGE_BRANCH` property to each `PublishBitbucketServerTag` task.

decoyjoe: I don't like publishing from a commit hash that doesn't exist, which is what happens when publishing from the transient merge commit.

My preference would be to disable publishing from PR builds, since we can't tag a specific commit. We can't tag the source branch commit because the transient merge commit may contain changes that aren't reflected when building the source branch on its own.

To publish and tag during a PR, we just need to ensure AppVeyor and Jenkins have feature branch builds still enabled when a PR is open. That branch build runs alongside the PR build and is the build that does the publishing of a prerelease package because it has a commit it can tag.

splatteredbits: I think there are reasons/problems with running both branch an PR builds at the same time, but can't remember off the top of my head.

splatteredbits: I don't see it as publishing from a commit hash that doesn't exist. It's the build that publishes, not the commit. That build does a merge and creates a new commit. The merge is just another build action, like linting, running a compiler, etc., just a step the build server is taking instead of explicitly doing it ourselves. If the build server didn't do the merge commit, and we had to implement it ourselves, it would be a `git merge` build step and Whiskey would use the branch and commit of the source branch in its build metadata.

decoyjoe: Here's the issue I see through an example: `branch` is using version 1.1.0 of a library. `main` is ahead and is using version 1.2.0 of that library. The newer version of the library on `main` 1.2.0 has a bug.

During the PR build, `main` is merged into `branch` and thus branch is now using the newer bugged library. It publishes an artifact and tags the `branch` commit.

Someone uses the `branch` build and encounters the bugged behavior. They find that the bug is associated with version of 1.2.0 of the internal library. So they lookup the commit that was tagged with that specific build from the PR. When they look at that tagged `branch` commit they find it's pinned to version 1.1.0 of the library, not 1.2.0.

Begin the confusion on why the `branch` build is using library version 1.2.0 when it's actually pinned to 1.1.0.

If we're going to be publishing from PR builds, which are merge commits of two branches, we probably need a way to identify that build as such. The PR build tag must indicate that it was built from a merge of `branch` and `main`, and not solely the `branch` commit.

splatteredbits: If `branch` is pinned to `1.1.0` but `main`, the target branch, is pinned to `1.2.0`, wouldn't that mean someone changed the dependency in `main` and that `branch` doesn't have that change yet? It would cause confusion, but the underlying problem isn't with the build and committed tag, it's that developers haven't updated `branch` with the changes on `1.2.0`.

And they couldn't even begin to discover this if Whiskey hadn't tagged the source branch! So always using the source branch would actually help in this situation.

I don't know that I've ever cared or wanted to know if a particular package was built from a PR merge commit, hence I don't think Whiskey needs to expose that information in anyway.

decoyjoe: > If branch is pinned to 1.1.0 but main, the target branch, is pinned to 1.2.0, wouldn't that mean someone changed the dependency in main and that branch doesn't have that change yet? It would cause confusion, but the underlying problem isn't with the build and committed tag, it's that developers haven't updated branch with the changes on 1.2.0.

Then perhaps the solution is to disallow publishing if the target branch contains commits that are not on the source branch. Only then are you sure that the branch build is using the same code as the PR build.

I personally place a lot of value on reproducible builds. Ideally I should be able to rebuild a commit and get a bit-for-bit identical build. It's hard to achieve but I tend to optimize towards that direction when I can.

splatteredbits: > Then perhaps the solution is to disallow publishing if the target branch contains commits that are not on the source branch. Only then are you sure that the branch build is using the same code as the PR build.

This is a policy that can and should be enforced by the build server and the source code tool, not Whiskey. GitHub has these requirements. I don't think

> I personally place a lot of value on reproducible builds. Ideally I should be able to rebuild a commit and get a bit-for-bit identical build. It's hard to achieve but I tend to optimize towards that direction when I can.

I agree but PR builds can be, by their very nature, not reproducible. I think that's a problem with the concept of PR builds that we shouldn't try to fix in Whiskey. I think this is higher-level problem than what commands get run in a build. We have developers reporting problems when a PR build doesn't publish a build. It is developers' responsibility to know that when they create a PR, those builds use code from two branches, and to build and publish when that makes sense. Whiskey shouldn't get in the way.

splatteredbits (approved):

